PR #3, opened 2019-04-22
`setup.py` improvements

butuzov:
Hi Ram, you have a really nice package here. I hope you don't mind if I change `setup.py` a bit? 

Changes are next:

+ find_packages with `exclude` argument  instead of complicated comprehension
+ only `setup` and `find_packages` imported from `setuptools`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,13 +1,9 @@
 # Copyright 2019 Ram Rachum.
 # This program is distributed under the MIT license.
 
-import setuptools
+from setuptools import find_packages, setup
 
-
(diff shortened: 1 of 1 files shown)


cool-RR:
Thanks for contributing! I like the `exclude`, but I don't like the `from` import. I prefer `import setuptools` and then using dotted syntax for the rest. Mind amending your pull request?

butuzov:
No problem with that.

cheers.

cool-RR:
Thanks Oleg!

PR #22, opened 2019-04-23
Use common practice approaches with requirements and packaging

bittner:
These changes add some things that are common practice for requirements management and packaging:

* Include requirements in test requirements file
* Include lax requirements for installing (strict requirements cause installation problems)
* Generate strict requirements (to document the last working state)
* Make reading files DRY in setup.py
* Add more trove classifiers

Note that we may need to add a `MANIFEST.in` file if releasing doesn't prove to work.

Also, we need `tests_require` only when running tests with `python setup.py test`, which we don't do.


bittner:
If you're interested we may also add a `Pipfile` using [pipenv](https://pipenv.readthedocs.io/en/latest/). This would encapsulate requirements for development and releasing in a single file.

It would look like follows, and would be generated by `pipenv`:

```toml
[[source]]
name = "pypi"
url = "https://pypi.org/simple"
verify_ssl = true

[dev-packages]
pip-tools = "*"
pytest = "*"
python_toolbox = "*"

[packages]
decorator = "==4.4.0"
future = "==0.17.1"
six = "==1.12.0"
```

I would also suggest to add and up-to-date check (= automatic generation of the requirements + comparison with committed state) to the CI pipeline. Then updating requirements gets never forgotten.

cool-RR:
Thanks! I like most of these changes.

Here are a couple of issues I've noticed:

1. The pip requirements now don't include version numbers. What if someone has ancient versions of our dependencies that won't have what we need?

2. I like your `read_file` function, but it seems that it'll feed one long string into `requires` etc. instead of a list of strings. Will that work?

Maybe before submitting your next changes, you could test uploading a version to test.pypi.org .

cool-RR:
I'm a little reluctant about using `pipenv`. I heard about it a lot, but I think it's not common practice yet. I'd rather stick to something more simple.

bittner:
> 1. The pip requirements now don't include version numbers. What if someone has ancient versions of our dependencies that won't have what we need?

If you want to go _really_ safe then use `>=` and the like. The bottom line is, specifying "too much" has proven commonly to cause more harm than good.

> 2. I like your `read_file` function, but it seems that it'll feed one long string into `requires` etc. instead of a list of strings. Will that work?

It works. I have consciously omitted using `split` or `splitlines`.

bittner:
Side-comment: You can adapt the feature branch to your liking. It's in this repo, and I don't mind if you modify my contribution.

Issue #28, opened 2019-04-23
Feature request: Print line containing actual function name for "call" event


czietz:
When the function decorator `@pysnooper.snoop()` is in a separate line, the "call" event will print that line instead of the one containing the actual function name. This can be seen in the example in README.md:
https://github.com/cool-RR/PySnooper/blob/master/README.md#example

While this is technically correct (for Python the function definition starts with the decorator), it would make traces easier to read if the actual function name was printed. Maybe a special case could be added for "call" events: if the source line reported by Python only contains the decorator, print the next source line instead.

cool-RR:
Agreed. This seems to be a weakness in Python's introspection ability. I'll be happy to look at a pull request, except I'll need to be reasonably sure it won't backfire in other cases.

czietz:
Since Python syntax (https://docs.python.org/3/reference/compound_stmts.html#function-definitions) stipulates that each decorator must end with a new line and, therefore, the actual function name cannot be in the same line, it should be rather easy to skip decorator lines without accidentally skipping too much.

I'll send a pull request later today.

cool-RR:
Great, and I'll want tests too. (Feel free to amend the existing tests, if
you'll need help let me know.)

PR #29, opened 2019-04-23
Print line containing actual function name for "call" event

czietz:
Implements https://github.com/cool-RR/PySnooper/issues/28, including tests.

### DIFF
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -183,12 +183,19 @@ class Tracer:
         now_string = datetime_module.datetime.now().time().isoformat()
         line_no = frame.f_lineno
         source_line = get_source_from_frame(frame)[line_no - 1]
+
         if event == 'call':
             # Skip lines containing function decorators to print actual
             # function name
             while source_line.lstrip()[0] == '@':
                 line_no += 1
                 source_line = get_source_from_frame(frame)[line_no - 1]
+            # Check that source_line is actually a function definition,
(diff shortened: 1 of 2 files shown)


cool-RR:
That's nice, but I'm a little worried about line_no going forward like that. I suggest that you do a search for a line that matches the expression `"^ +def" after any number of lines that start with "@". If you don't find such a line, it's possible we're dealing with a case we don't understand, so cancel your transformation and show the original line with the decorator, just to play it safe.

If you can also add a test for that safe behavior, that'll be extra cool, but that's optional.

czietz:
I added a check that the line that is found actually starts with `def` (after possible white-space, of course). 

After looking at the grammar for a Python function definition (https://docs.python.org/3/reference/compound_stmts.html#function-definitions), I fail to come up with _any_ test-case that could trigger this. If you have an idea, I'll be happy to implement the test.

czietz:
I fixed items 1-4. 

I have no idea how to write a test for 3, though. I tested that it works as intended by locally modifying the source to search for `xef` instead of `def`, in which case it will correctly revert to the original line, i.e., the decorator. But that's nothing that can be added to a test-case.

If you insist on a test-case for 3, then please consider this pull-request retracted.

cool-RR:
Great work @czietz ! I merged your commits, did some refactoring on them and added a test for lambda. (I merged them directly into the repo, not through the pull request, so I had to close it.)

PR #31, opened 2019-04-23
Add support for generators, requires a flag when snooping

### LINKED ISSUE #9: Feature request: Support generators
For example:

```Python
from __future__ import print_function
import pysnooper

@pysnooper.snoop()
def fib(max_n):
    a, b = 0, 1
    while a < max_n:
        yield a
        a, b = b, a + b

def main():
    for n in fib(4):
        print(n)

if __name__ == '__main__':
    main()
```

Output:

```Bash
$ python mytest.py 
0
1
1
2
3
```

knixeur:
Due to the nature of Python trating a method as a generator when a yield
keyword is defined, it's hard to make this dynamic.
I ended up adding an additional parameter is_generator which needs to
be set to True when snooping a generator.

An additional decorator could be defined to make it easier, like @pysnooper.snoopgen which could call snoop with is_generator=True

Fixes #9

### DIFF
--- a/pysnooper/pysnooper.py
+++ b/pysnooper/pysnooper.py
@@ -33,8 +33,7 @@ def get_write_function(output):
     return write
 
 
-
-def snoop(output=None, variables=(), depth=1, prefix=''):
+def snoop(output=None, variables=(), depth=1, prefix='', is_generator=False):
     '''
     Snoop on the function, writing everything it's doing to stderr.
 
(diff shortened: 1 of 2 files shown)


cool-RR:
1. Use `inspect.isgeneratorfunction` to get rid of your generator argument.
2. The context manager should be held when doing `next`. i.e. tracing should be active only when the generator is running. So instead of a for loop, you need a `while True` loop that does `next` until a `StopIteration` is thrown. It should acquire the tracer on every `next`.

knixeur:
1. I tried that, but since Python find a yield statement, it transforms always returns a generator. Feel free to patch this PR and/or base a better one from this.
2. I'll do that, I was actually using yield from when I did these changes but it's not supported in Python 2.7.

knixeur:
Sorry, I won't have to finish this. I tried quickly to apply change 2 but it's giving me some headaches, I'll close this PR and if I have some time in the following weeks I'll give it another chance.

cool-RR:
I understand. Thanks for working on this!

PR #38, opened 2019-04-24
python2 collections

cc20101001:
python2 collections

### DIFF
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -5,6 +5,10 @@ import types
 import sys
 import re
 import collections
+if sys.version.startswith("3"):
+    from collections import ChainMap
+else:
+    from ConfigParser import _Chainmap as ChainMap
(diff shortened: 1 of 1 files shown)


cool-RR:
Thanks @cc20101001 ! Great job!

pysnooper/tracer.py
```diff
@@ -5,6 +5,10 @@
 import sys
 import re
 import collections
+if sys.version.startswith("3"):
+    from collections import ChainMap
+else:
+    from ConfigParser import _Chainmap as ChainMap
```
bittner: The pythonic way would be using `try` + `except` instead of probing for the Python version.

```python
try:
    from ConfigParser import _Chainmap as ChainMap
except ImportError:
    from collections import ChainMap
```

cool-RR: Correct, I fixed it later :)

PR #45, opened 2019-04-24
Trying to fix Travis problem

cool-RR:


bittner:
Can we delete the `development` branch?

cool-RR:
Hmm, I use it often for experiments, so I prefer to keep it.

Issue #1, opened 2019-04-21
Feature request: Add output coloring, option to disable


cool-RR:

lt94:
Hey, i'd like to work on this one,any good advice for me?

cool-RR:
Great!

I think this isn't too tricky, just find the color characters for the
shell, and use different colors for different parts of the output. If
there's anything I'd want to correct about your pull request, I'll tell
you.

lt94:
I hava already fix this one,you can see my last commit [#1](https://github.com/cool-RR/PySnooper/compare/master...lt94:master), but got no idea about how to test color. would you mind giving me some help,thx~

cool-RR:
Great, can you submit a pull request to this repo? I tried making sense of your code but couldn't separate your commits from mine on your repo.

lt94:
i made the pr~~O(∩_∩)O

cool-RR:
This PR is up for grabs again. It's a good one, I hope someone takes it.

Mark-Jung:
I'll give it a shot.

AtHeartEngineer:
What about doing this with pycolorama? @Mark-Jung lemme know if you want help, definitely looking forward to this PR.

Mark-Jung:
I haven't looked into this yet, but I will asap. Have been swamped with work this week.

alexmojaki:
Before jumping into a PR with colorama you should be aware that there is a reluctance here to use any dependencies, see #48 and #68

cool-RR:
@AtHeartEngineer : Alex is correct, I prefer to avoid dependencies for this project.

alexmojaki:
> It could be an optional dependency, if you want color

I know @cool-RR is not interested in maintaining optional dependencies either, I tried that too in #71

I tried to make this project extensible in #103 so that another project could be built on top of it with extra features through dependencies, but that's been rejected. I'd already written some code using Pygments which added syntax highlighting in the terminal.

cool-RR:
Anyone reading this: This feature is up for grabs for anyone interested in implementing it.

alexmojaki:
[My version of PySnooper](https://github.com/alexmojaki/snoop) is out and it is colorful! I'd love to hear people's thoughts about the design and use of color [here](https://github.com/alexmojaki/snoop/issues/1).

[...]

PR #47, opened 2019-04-25
Fixs #1 Add output coloring, option to disable

lt94:


cool-RR:
I see a bunch of problems with this PR:
1. It seems like you added some kind of framework for coloring, but didn't really add the logic of coloring different parts of the output in different colors. That's kind of the point.
2. It's not clear why you have 3 commits. I think it could be merged to one, if you want several commits the difference between them should be emphasized in the message.
3. You need to rebase on the latest changes.
4. You included whitespace changes to unrelated files, don't do that.

lt94:
i will close this pr and correct the problems~~Thanks a lot

cool-RR:
No need to close, you could update it later, no rush.

PR #50, opened 2019-04-25
Get write method directly without ABCs, MROs, etc.

alexmojaki:
Perhaps I'm missing something, but it seems like all the WritableStream-related code was an unnecessarily over-complicated way to check if an attribute was present. If it quacks like a duck and has a write attribute, isn't that good enough?


cool-RR:
Correct, it was a bit redundant. But I like that code, sorry.

PR #51, opened 2019-04-25
Fixs #1

lt94:


cool-RR:
I'm gonna be gentle because it's a draft, but: You know we want to not just color the whole output in one color, but give different colors to different parts of the output? That's the entire point.

PR #52, opened 2019-04-25
Basic fixes for linters

alexmojaki:
Remove unused variables and imports, plus whitespace and other style improvements for PEP8

### DIFF
--- a/tests/test_pysnooper.py
+++ b/tests/test_pysnooper.py
@@ -2,27 +2,25 @@
 # This program is distributed under the MIT license.
 
 import io
-import re
-import abc
 
-from python_toolbox import caching
 from python_toolbox import sys_tools
 from python_toolbox import temp_file_tools
-from pysnooper.third_party import six
 
 import pysnooper
-
+from pysnooper.third_party import six
 from .utils import (assert_output, VariableEntry, CallEntry, LineEntry,
-                    ReturnEntry, OpcodeEntry, ReturnValueEntry, ExceptionEntry)
+                    ReturnEntry, ReturnValueEntry)
 
 
 def test_string_io():
     string_io = io.StringIO()
+
     @pysnooper.snoop(string_io)
     def my_function(foo):
         x = 7
         y = 8
         return y + x
+
     result = my_function('baba')
     assert result == 15
     output = string_io.getvalue()
@@ -41,6 +39,7 @@ def test_string_io():
         )
     )
 
+
 def test_variables():
 
     class Foo(object):
@@ -50,7 +49,7 @@ def test_variables():
         def square(self):
             self.x **= 2
 
-    @pysnooper.snoop(variables=('foo.x', 're'))
+    @pysnooper.snoop(variables=('foo.x', 'io'))
     def my_function():
         foo = Foo()
         for i in range(2):
@@ -83,6 +82,7 @@ def test_variables():
             ReturnValueEntry('None')
         )
     )
+
 
 def test_depth():
     string_io = io.StringIO()
@@ -144,7 +144,6 @@ def test_depth():
 
 
 def test_method_and_prefix():
-
     class Baz(object):
         def __init__(self):
             self.x = 2
@@ -178,15 +177,17 @@ def test_method_and_prefix():
         prefix='ZZZ'
     )
 
-def test_file_output():
 
+def test_file_output():
     with temp_file_tools.create_temp_folder(prefix='pysnooper') as folder:
         path = folder / 'foo.log'
+
         @pysnooper.snoop(str(path))
-        def my_function(foo):
+        def my_function(_foo):
             x = 7
             y = 8
             return y + x
+
         result = my_function('baba')
         assert result == 15
         with path.open() as output_file:
@@ -194,8 +195,8 @@ def test_file_output():
         assert_output(
             output,
             (
-                VariableEntry('foo', value_regex="u?'baba'"),
-                CallEntry('def my_function(foo):'),
+                VariableEntry('_foo', value_regex="u?'baba'"),
+                CallEntry('def my_function(_foo):'),
                 LineEntry('x = 7'),
                 VariableEntry('x', '7'),
                 LineEntry('y = 8'),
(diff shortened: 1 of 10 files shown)


tests/test_pysnooper.py
```diff
@@ -50,7 +49,7 @@ def __init__(self):
         def square(self):
             self.x **= 2
 
-    @pysnooper.snoop(variables=('foo.x', 're'))
+    @pysnooper.snoop(variables=('foo.x', 'io'))
```
alexmojaki: Test was failing since I removed the unused `import re`

tests/test_pysnooper.py
```diff
@@ -178,24 +177,26 @@ def square(self):
...
 
-def test_file_output():
 
+def test_file_output():
     with temp_file_tools.create_temp_folder(prefix='pysnooper') as folder:
         path = folder / 'foo.log'
+
         @pysnooper.snoop(str(path))
-        def my_function(foo):
+        def my_function(_foo):
```
alexmojaki: At least to my IDE, the leading underscore indicates an unused variable so it doesn't show a warning about being unused.

alexmojaki:
On this note, you have checks in tox and travis for linting, but of course there are tons of linter errors so the checks are allowed to fail. What's the goal? If something was measuring the number of errors and showing the change caused by a commit or PR I would understand, but I don't see anything like that.

bittner:
You may want to ignore `pysnooper/third_party` for both Flake8 and Pylint in [tox.ini](https://github.com/cool-RR/PySnooper/blob/master/tox.ini). This will reduce the area we have to put work in, and we may get a complaint-free lint run, eventually.

Our goal should be to remove the [allowed_failures](https://github.com/cool-RR/PySnooper/blob/master/.travis.yml#L25-L28), sooner or later.

cool-RR:
@alexmojaki Good work. I merged two of your commits directly, please rebase this pull request.

cool-RR:
@alexmojaki I disagree with much of PEP8 and many of the things that PyLint reports.

If you'll find my approach too idiosyncratic and lose interest in contributing to this part, I'd understand! Your effort is appreciated regardless of whether we'll agree.

In general, I think that linting tools can often cause more harm than good, because they can increase frustration. Also, people sometimes put "noqa" or other nonsense on code to mark certain sections as not requiring linting, which is really annoying in my opinion. Linters should serve us, not the other way around.

Looking at your changes, I agree with: 
1. One space before comments.
2. Removing unused imports. (Except for cases like `OpcodeEntry` and `ExceptionEntry`.)
3. Removing unused variables.
4. Fixing spacing between lines, mostly.

I disagree with: 
1. Enforcing double quotes for multiline strings. I prefer single quotes.
2. Undoing the weird indentation I like to do, like in `result = {key: get_shortish_repr(value) for key, value ...`

bittner:
Ram, just to make sure you get this discussion in the right throat, a bit of history: I've had more than enough of discussions on the topic of using linters to enforce code quality. It always boils down to explaining the same things over again: It's about a common understanding on doing software development (code co-creation). It is common to call this "best practices".

Specifically, on your objections:

- I do agree on `# noqa` being wrong. The right way is to tune the linter configuration ti the type of your project. For generic Python the defaults of Flake8 and Pylint are actually quite good. You need to get used to it, though (if you haven't worked with especially Pylint before, who is pretty strict -- but cool).
- Double-quotes on docstrings is proposed by [PEP 257](https://www.python.org/dev/peps/pep-0257/). It's a good idea to follow PEPs. (Python is strong for this! Best example: PEP 20)
- Indentation is arguable, but almost always if not in every single case Flake8 is right. Sometimes there are different ways to structure code to make the indentation complaint go away. It is fine to find different solutions (for splitting up lines) in different parts of the code. But why not please Flake8? It's a tool used by probably all serious Python developers. 

Bottom line:

- It's best to start with linting from commit one.
- When introduced later (as with us) it causes pain. But that pains goes away once maintainers (and contributors) get used to it. The satisfaction of learning something when linters catch a violation is quite nice, every now and then.
- That educating contributors to produce good code scares off valuable contributions is a widespread myth. Think: If it is bad quality, is it maybe a questionable contribution?

Hope that helps! :+1:

alexmojaki:
I rebased, but I'm not sure if I did something wrong seeing all these other commits showing up here. I've never had a solid grasp of rebasing (as opposed to merging). I also haven't had to fetch from a source repo before to sync my fork. If it looks like I did something wrong, you can see my shell session [here](https://pastebin.com/xyALCQvu).

If you want to avoid people making specific stylistic changes, I suggest that you document your preferences and disable the appropriate warnings when the code is checked by tox (and thus travis). Right now anyone looking at this repo is probably going to assume that the desire is to make it fully PEP8 compliant.

I don't have strong feelings about quotes or indentations. I just don't like my IDE (PyCharm) showing a bunch of warnings that can easily be removed. It's distracting, annoying, and makes it harder to spot the new warnings that I need to actually pay attention to. I can disable those warnings, but I didn't because I assumed that everyone wanted to get rid of them. I also don't think it seems nice to make every potential contributor go through that process of configuring their own tools, so you might want to provide or request configurations files. Config files for flake8, pylint, and bandit would kill two birds with one stone.

cool-RR:
> I rebased, but I'm not sure if I did something wrong

You indeed rebased wrong. I looked at your terminal dump, but it's hard to figure out, and I'm not the best at rebasing either. 

I picked your commit manually now, added my opinionated edits to it, merged it to master and made a release.

Thanks for your contribution! 

> I suggest that you document your preferences and disable the appropriate warnings when the code is checked by tox (and thus travis)

To be honest I don't really care about PyLint checks, this is more Peter's thing. @bittner , if you're willing to add an ignore to the rules that I said I don't agree with, that'll be helpful. If you'll lose interest because of my idiosyncratic position, I would totally understand, and if that's the case it's probably best to just remove the pylint/flake8 altogether.

> I don't have strong feelings about quotes or indentations. I just don't like my IDE (PyCharm) showing a bunch of warnings that can easily be removed.

I do have a strong opinion about that. I give classes where I teach people Python programming with PyCharm, and one of the common mistakes is giving too many fucks about PyCharm's warnings. 

PyCharm works for us, not the other way around. Every project has different styling choices. If PyCharm warnings help you catch real-life errors, or help guide you to a style that you prefer, that's good. But when you reach the point where you're changing code just to remove a PyCharm warning, that's when you know you've gone too far. Or you're in a corporate environment, where hopefully you're getting paid handsomely for having gone too far :)

Thank you for your contribution Alex!

PR #56, opened 2019-04-25
Allow tracking arbitrary expressions in 'variables' using eval

alexmojaki:
I'm keen to hear suggestions/objections.

### DIFF
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -4,10 +4,6 @@
 import sys
 import re
 import collections
-try:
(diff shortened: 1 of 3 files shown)


cool-RR:
Das ist cool. Add a test that does a couple of interesting cases, like dict access and function calls.

alexmojaki:
Dang, I really need to read up on rebasing.

cool-RR:
Haha, you really should. Great worked, I merged it directly. You also implemented #58 on the way.

PR #62, opened 2019-04-26
Support color in output #1

lt94:
I'm so sorry that i get wrong idea about outoup coloring. So i make a new pr,which result like the picture below:

Show output coloring, default color dict {"newish":"green","modified":"orange","def":"blue","return":"cyan"}::
        
```python
@pysnooper.snoop()
```

to disabled colorful output, just make color to *None*:

```python
@pysnooper.snoop(color=None)
``` 

or just change centain color:

```python
@pysnooper.snoop(color={"newish":"red"})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,24 @@ Start all snoop lines with a prefix, to grep for them easily:
 @pysnooper.snoop(prefix='ZZZ ')
 ```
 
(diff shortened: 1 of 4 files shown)


cool-RR:
I'll look at this. Where did you take `hue.py` from?

cool-RR:
This implementation is so-so, I'm sorry. I'll have to change so many things in it that it's not worthwhile for me to merge. Maybe you should pick an easier task.

Issue #19, opened 2019-04-23
Compatibility with awsebcli


lodb:
There are requirement incompatibilities with the `awsebcli` package from amazon. I'm not sure if this has an easy fix or pysnooper _really_ needs the specified versions of `future` and `six`:

`awsebcli 3.15.0 has requirement future<0.17.0,>=0.16.0, but you'll have future 0.17.1 which is incompatible.`
`awsebcli 3.15.0 has requirement six<1.12.0,>=1.11.0, but you'll have six 1.12.0 which is incompatible.`

cool-RR:
I understand the problem. I don't think it's worthwhile to fix on PySnooper's end, because awsebcli is insisting on older versions, and other packages might insist on even older versions.

What you can do is install PySnooper manually by downloading and running `python setup.py install`. That'll likely work.

PR #68, opened 2019-04-27
Use cheap_repr

alexmojaki:
[cheap_repr](https://github.com/alexmojaki/cheap_repr) is my library. It may seem obscure but it's a core dependency of my debugger [birdseye](https://github.com/alexmojaki/birdseye), so it's battle tested.

Using it would be a significant improvement to performance and provide more useful information to the user.

It looks like you don't want to have external dependencies, which I'm guessing is going to be a problem. What's that about?

### DIFF
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -9,27 +9,18 @@ import itertools
 
 from .third_party import six
 
+from cheap_repr import cheap_repr
+
 MAX_VARIABLE_LENGTH = 100
 ipython_filename_pattern = re.compile('^<ipython-input-([0-9]+)-.*>$')
 
 
-def get_shortish_repr(item):
-    try:
-        r = repr(item)
-    except Exception:
-        r = 'REPR FAILED'
-    r = r.replace('\r', '').replace('\n', '')
-    if len(r) > MAX_VARIABLE_LENGTH:
-        r = '{truncated_r}...'.format(truncated_r=r[:MAX_VARIABLE_LENGTH])
-    return r
-
-
 def get_local_reprs(frame, variables=()):
-    result = {key: get_shortish_repr(value) for key, value
+    result = {key: cheap_repr(value) for key, value
                                                      in frame.f_locals.items()}
     for variable in variables:
         try:
-            result[variable] = get_shortish_repr(
+            result[variable] = cheap_repr(
                 eval(variable, frame.f_globals, frame.f_locals)
             )
         except Exception:
@@ -230,7 +221,7 @@ class Tracer:
                    '{line_no:4} {source_line}'.format(**locals()))
 
         if event == 'return':
-            return_value_repr = get_shortish_repr(arg)
+            return_value_repr = cheap_repr(arg)
             self.write('{indent}Return value:.. {return_value_repr}'.
                                                             format(**locals()))
 
--- a/requirements.in
+++ b/requirements.in
@@ -1,1 +1,1 @@
-# That's right baby! No dependencies!
+cheap_repr
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,5 @@
 #
 #    pip-compile --output-file requirements.txt requirements.in
 #
-cheap-repr==0.3.0
-future==0.17.1            # via cheap-repr
+cheap-repr==0.3.1
 qualname==0.1.0           # via cheap-repr
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,6 @@
 #
 #    pip-compile --output-file requirements.txt requirements.in
 #
+cheap-repr==0.3.0
+future==0.17.1            # via cheap-repr
+qualname==0.1.0           # via cheap-repr
--- a/tox.ini
+++ b/tox.ini
@@ -16,6 +16,7 @@ description = Unit tests
 deps =
     pytest
     python_toolbox
+    -rrequirements.txt
 commands = pytest
 setenv =
     # until python_toolbox is fixed


cool-RR:
Sorry, this looks like it's solving a problem I don't have, adding a dependency I'd rather avoid, and possibly creating new problems on the way.

alexmojaki:
Oh, I definitely think it solves problems. Let's start with performance. Here's a benchmark:

```python
from io import StringIO

from littleutils import timer

from pysnooper import snoop


def find_primes(limit):
    primes = [2]
    for candidate in range(3, limit + 1, 2):
        for p in primes:
            if p * p > candidate:
                primes.append(candidate)
                break

            if candidate % p == 0:
                break

    return primes


@snoop(StringIO())
def is_prime(x):
    primes = find_primes(int(x ** 0.5))
    for p in primes:
        if x % p == 0:
            return False
    return True


with timer():
    print(is_prime(22801763489))
```

Without snooping, this takes about 0.2 seconds.

With snooping with cheap_repr, it takes about 5 seconds.

With snooping on master, it takes about 60 seconds.

alexmojaki:
A more serious 'performance' problem would be if you had some Django code that started like this:

```python
queryset = BigTable.objects.all()
# add some dynamic filters or something later
```

`snoop` evaluating the `repr` of that would cause the entire table to be downloaded into memory, which may run out.

In terms of usefulness, consider this code:

```python
import requests
from pysnooper import snoop

@snoop()
def main():
    response = requests.get('https://pypi.org/pypi/PySnooper/json').json()

main()
```

The current version shows something like this:

```
New var:....... response = {'urls': [{'packagetype': 'bdist_wheel', 'md5_digest': 'feb064a5c883f2f33cc401cb1dbd6084', 'requires...
```

which contains almost no useful information. With cheap_repr, it shows:

```
New var:....... response = {'urls': [{'url': 'https://files.pythonhosted.o...r-0.0.22-py2.py3-none-any.whl', 'upload_time': '2019-04-26T09:48:42', 'python_version': 'py2.py3', 'requires_python': None, ...}, {'url': 'https://files.pythonhosted.o...f3b24/PySnooper-0.0.22.tar.gz', 'upload_time': '2019-04-26T09:48:44', 'python_version': 'source', 'requires_python': None, ...}], 'info': {'project_url': 'https://pypi.org/project/PySnooper/', 'license': '', 'home_page': 'https://github.com/cool-RR/PySnooper', 'classifiers': ['Environment :: Console', 'Intended Audience :: Developers', 'License :: OSI Approved :: MIT License', ..., 'Programming Language :: Python :: Implementation :: CPython', 'Programming Language :: Python :: Implementation :: PyPy', 'Topic :: Software Development :: Debuggers'], ...}, 'releases': {'0.0.4': [{...}, {...}], '0.0.9': [{...}, {...}], '0.0.13': [{...}, {...}], '0.0.7': [{...}, {...}], ...}, 'last_serial': 5191977}
```

which hints at what information is there and how to get it.

cool-RR:
It's nice, but not worth adding a dependency in my opinion, sorry. I think that when people use PySnooper, they don't care all that much about performance, it's just a debugging run to find a problem, and then they remove the decorator. Regarding the big reprs, that's nice but it's a small enough concern that until we have users who complain about it, I think it's not worth fixing.

alexmojaki:
Right, so this is why I asked at the very beginning, why the reluctance to have dependencies?

Why would performance not be an issue when casually debugging? You run some code, it's almost instant, you add snoop to find why it's not doing the right thing, and you have to wait 60 seconds or worse. How is that not terrible for user experience?

Why limit the use cases to temporary debugging? This could very easily be used as a lazy sort of logging in deployed (but hopefully not critical) code in anticipation of problems not yet found. [hunter](https://github.com/ionelmc/python-hunter) is a similar library which suggests being used that way.

cool-RR:
> Right, so this is why I asked at the very beginning, why the reluctance to have dependencies?

Dependencies can cause problems, for example version mismatches, as one user reported last week. These problems can usually be solved, but if they can be completely avoided, that's a big plus.

> Why would performance not be an issue when casually debugging? You run some code, it's almost instant, you add snoop to find why it's not doing the right thing, and you have to wait 60 seconds or worse.

I think that the 60 second example was a contrived one. I think that most scenarios would be slowed down by a bit, but not that much. When using debuggers, people accept that their code will run much slower, and that's okay.

> Why limit the use cases to temporary debugging? This could very easily be used as a lazy sort of logging

Sounds dangerous to me. It's not inconceivable that people will use it this way successfully. But I don't want to make concessions for a use case which I'm not advocating.

alexmojaki:
I assume you mean [this](https://github.com/cool-RR/PySnooper/issues/19) issue about version mismatches. That wouldn't happen any more because you [stopped requiring specific versions](https://github.com/cool-RR/PySnooper/commit/4a0bec9ded9124a6de5417143466f6f3f6220012#diff-a24204f7139342bdf2a7d7d13ca4dab0), which is a good thing. As the reporter said, you probably don't need the latest version of six.

`cheap_repr` itself isn't likely to cause version issues. [No other library on github](https://github.com/search?q=cheap_repr&type=Code) uses it except birdseye, which doesn't specify a version. It only has one tiny dependency (even transitively), qualname, for which it doesn't specify a version, and which hasn't been updated in 4 years.

So what other problems do you have with dependencies?

> I think that the 60 second example was a contrived one.

I created it for a demo, yes, but that's not the point. All it takes to hit the performance problem is the presence of one big variable. That's really not a very unlikely condition. The variable here isn't even that big, it's a list of about 14,000 numbers. If I make it contain 100,000 numbers, the code takes practically forever.

> When using debuggers, people accept that their code will run much slower, and that's okay.

I think people would accept their code taking 5 seconds, not 60. I would probably interrupt the code before it finished, conclude that I can't use pysnooper for this problem, and become more reluctant to use it in the future.

Besides, I don't get your point. Why live up to the expectation of poor performance when you can easily improve it?

cool-RR:
> That wouldn't happen any more

Yes, and other problems will happen. I had one happen last week with Travis. And then we'll solve them and other problems will happen.

PySnooper isn't my job, I'm doing this for fun in my spare time. I don't want to spend time in dependency hell.

> Why live up to the expectation of poor performance when you can easily improve it?

I think that if it's a major problem, people will report it. You say "easily" because it seems easy to you, because you're not the person responsible for dealing with any kind of dependency problems that'll prevent people from using this program. I am. I'm guessing you're not willing to guarantee that you'll personally fix any dependency problem that a person might have as a result of this change.

To be honest I hope that this discussion is going to end soon.

alexmojaki:
> I'm guessing you're not willing to guarantee that you'll personally fix any dependency problem that a person might have as a result of this change.

Sure I am, I'm happy to take responsibility for any problems I cause.

cool-RR:
So if two years from now, I have a problem because of the dependency, I can
call you up and you'll guarantee to solve the problem and compensate me for
any time lost because of the problem? Can I get it in writing with your
signature, ID and a security deposit? ;)

Unless the answer is yes, I'd like to close this pull request.

PR #71, opened 2019-04-28
Make cheap_repr an optional dependency

alexmojaki:
Compromise to #68. cheap_repr will only be activated if the user installs it separately, which we can recommend in the docs. We could make it an install extra in setup.py, but I don't think there's much point in that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ jobs:
   - { stage: lint, python: 3.7, env: TOXENV=pylint }
   - { stage: lint, python: 3.7, env: TOXENV=bandit }
   - { stage: lint, python: 3.7, env: TOXENV=readme }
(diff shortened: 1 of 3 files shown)


cool-RR:
Nice compromise, but then we'll have to add testing for it, and I don't
think it's worth it. It increases the complexity of the project.

Here's an idea for you: Add an option to cheap_repr to monkeypatch the
built-in repr method. Then it'll automatically take effect in PySnooper and
everywhere else.

alexmojaki:
It's easy to add testing, I just did it.

Monkeypatching the built in repr would be disastrous and no one should ever do that.

cool-RR:
> Monkeypatching the built in repr would be disastrous and no one should ever do that.

For debugging purposes, I think it's acceptable.

Sorry, I wouldn't like to maintain this code.

PR #72, opened 2019-04-28
Use reprlib

alexmojaki:
This isn't as good as cheap_repr, but it uses the standard library.

### DIFF
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -8,28 +8,30 @@ import datetime as datetime_module
 import itertools
 try:
     import reprlib
+    import builtins
 except ImportError:
     import repr as reprlib
+    import __builtin__ as builtins
 
 from .third_party import six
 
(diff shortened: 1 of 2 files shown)


cool-RR:
1. Will this have identical output for most simple objects? For example will the example from the readme be exactly the same?
2. Can you show how the results look when the output is indeed too long?
3. I see that the default for `maxother` is 20. That is quite short. Is there a way to raise that to 100 without changing anything globally?
4. Does this also catch arbitrary exceptions in the `__repr__`? Because you removed my `except`.
5. The Python 3 import should be before the Python 2 import, to show our respect for Python 3 :)

alexmojaki:
1. Yes
2. Added a test
3. Done. Users can configure it.
4. It did, but I've made extra sure while also making it more informative.
5. Done.

cool-RR:
Looking good. Please rename `MyRepr` to `Repr`. And of course, I assume you're working on the failed tests.

cool-RR:
Great work! I like that you added tests for the exception case.

Issue #73, opened 2019-04-28
Adding to list may not trigger modified var line


alexmojaki:
Example script:

```python
import pysnooper

@pysnooper.snoop()
def foo():
    x = []
    for i in range(10):
        x.append(i)

foo()
```

Output:

```
13:38:44.029228 call         5 def foo():
13:38:44.029456 line         6     x = []
New var:....... x = []
13:38:44.029540 line         7     for i in range(10):
New var:....... i = 0
13:38:44.029621 line         8         x.append(i)
Modified var:.. x = [0]
13:38:44.029688 line         7     for i in range(10):
Modified var:.. i = 1
13:38:44.029751 line         8         x.append(i)
Modified var:.. x = [0, 1]
13:38:44.029816 line         7     for i in range(10):
Modified var:.. i = 2
13:38:44.029880 line         8         x.append(i)
Modified var:.. x = [0, 1, 2]
13:38:44.029948 line         7     for i in range(10):
Modified var:.. i = 3
13:38:44.030015 line         8         x.append(i)
Modified var:.. x = [0, 1, 2, 3]
13:38:44.030083 line         7     for i in range(10):
Modified var:.. i = 4
13:38:44.030150 line         8         x.append(i)
Modified var:.. x = [0, 1, 2, 3, 4]
13:38:44.030220 line         7     for i in range(10):
Modified var:.. i = 5
13:38:44.030290 line         8         x.append(i)
Modified var:.. x = [0, 1, 2, 3, 4, 5]
13:38:44.030364 line         7     for i in range(10):
Modified var:.. i = 6
13:38:44.030436 line         8         x.append(i)
Modified var:.. x = [0, 1, 2, 3, 4, 5, ...]
13:38:44.030508 line         7     for i in range(10):
Modified var:.. i = 7
13:38:44.030580 line         8         x.append(i)
13:38:44.030638 line         7     for i in range(10):
Modified var:.. i = 8
13:38:44.030712 line         8         x.append(i)
13:38:44.030770 line         7     for i in range(10):
Modified var:.. i = 9
13:38:44.030843 line         8         x.append(i)
13:38:44.030901 line         7     for i in range(10):
13:38:44.030960 return       7     for i in range(10):
```

Partway through `x` seems to stop changing and the user might be misled into thinking it has, which could be a serious problem when trying to figure out a bug. Here it's obvious but in a more complex script it may not be.

Or suppose the user is building up a list of strings to make a document and does this:

[...]

PR #74, opened 2019-04-28
Add classes to automatically track attributes, keys, etc. of variables

### LINKED ISSUE #59: Feature request: Add `exploding_variables` argument
Any exploding variable will not only be tracked, but all its attributes will be tracked too.

alexmojaki:
Closes #59 

Sample usage:

```python
from types import SimpleNamespace

import pysnooper
from pysnooper.variables import Attrs, Indices, Keys, Enumerate


@pysnooper.snoop(variables=(
        Keys('d', exclude='c'),
        Attrs('point'),
        Indices('lst')[-3:],
        Enumerate('s'),
))
def foo():
    d = {'a': 1, 'b': 2, 'c': 'ignore'}
    point = SimpleNamespace(x=3, y=4)
    lst = list(range(1000))
    s = {7, 8, 9}


foo()
```

Output:

```
19:01:11.689207 call        13 def foo():
19:01:11.779673 line        14     d = {'a': 1, 'b': 2, 'c': 'ignore'}
New var:....... (d)['a'] = 1
New var:....... (d)['b'] = 2
New var:....... d = {'a': 1, 'b': 2, 'c': 'ignore'}
19:01:11.872593 line        15     point = SimpleNamespace(x=3, y=4)
New var:....... (point).x = 3
New var:....... (point).y = 4
New var:....... point = namespace(x=3, y=4)
19:01:12.000584 line        16     lst = list(range(1000))
New var:....... (lst)[997] = 997
New var:....... (lst)[998] = 998
New var:....... (lst)[999] = 999
New var:....... lst = [0, 1, 2, 3, 4, 5, ...]
19:01:12.128228 line        17     s = {7, 8, 9}
New var:....... (s)<0> = 8
New var:....... (s)<1> = 9
New var:....... (s)<2> = 7
New var:....... s = {7, 8, 9}
19:01:12.260525 return      17     s = {7, 8, 9}
Return value:.. None
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,14 @@ See values of some variables that aren't local variables:
 @pysnooper.snoop(variables=('foo.bar', 'self.whatever'))
 ```
 
+Expand values to see all their attributes or items of lists/dictionaries:
+
+```python
+@pysnooper.snoop(exploding_variables=('foo', 'self'))
+```
+
+(see [Advanced Usage](#advanced-usage) for more control)
+
 Show snoop lines for functions that your function calls:
 
 ```python
@@ -98,6 +106,24 @@ Start all snoop lines with a prefix, to grep for them easily:
 ```console
 $ pip install pysnooper
 ```
+
+# Advanced Usage #
+
+`exploding_variables` will automatically guess how to expand the expression passed to it based on its class. You can be more specific by using one of the following classes:
+
+```python
+import pysnooper
+
+@pysnooper.snoop(variables=(
+    pysnooper.Attrs('x'),    # attributes
+    pysnooper.Keys('y'),     # mapping (e.g. dict) items
+    pysnooper.Indices('z'),  # sequence (e.g. list/tuple) items
+))
+```
+
+Exclude specific keys/attributes/indices with the `exclude` parameter, e.g. `Attrs('x', exclude=('_foo', '_bar'))`.
+
+Add a slice after `Indices` to only see the values within that slice, e.g. `Indices('z')[-3:]`.
 
 # Contribute #
 
--- a/README.md
+++ b/README.md
@@ -75,10 +75,10 @@ If stderr is not easily accessible for you, you can redirect the output to a fil
 
 You can also pass a stream or a callable instead, and they'll be used.
 
-See values of some variables that aren't local variables:
+See values of some expressions that aren't local variables:
 
 ```python
-@pysnooper.snoop(variables=('foo.bar', 'self.whatever'))
+@pysnooper.snoop(watch=('foo.bar', 'self.x["whatever"]'))
 ```
 
 Expand values to see all their attributes or items of lists/dictionaries:
--- a/pysnooper/__init__.py
+++ b/pysnooper/__init__.py
@@ -2,6 +2,7 @@
 # This program is distributed under the MIT license.
 
 from .pysnooper import snoop
+from .variables import Attrs, Exploded, Indices, Keys
 import collections
 
 __VersionInfo = collections.namedtuple('VersionInfo',
--- a/pysnooper/__init__.py
+++ b/pysnooper/__init__.py
@@ -2,13 +2,13 @@
 # This program is distributed under the MIT license.
 
 from .pysnooper import snoop
-from .variables import Attrs, Exploded, Indices, Keys
+from .variables import Attrs, Exploding, Indices, Keys
 import collections
 
 __VersionInfo = collections.namedtuple('VersionInfo',
                                        ('major', 'minor', 'micro'))
 
-__version__ = '0.0.26'
+__version__ = '0.0.27'
 __version_info__ = __VersionInfo(*(map(int, __version__.split('.'))))
 
 del collections, __VersionInfo # Avoid polluting the namespace
--- a/pysnooper/pysnooper.py
+++ b/pysnooper/pysnooper.py
@@ -39,7 +39,8 @@ def get_write_and_truncate_functions(output):
     return (write, truncate)
 
 
-def snoop(output=None, variables=(), exploded_variables=(), depth=1, prefix='', overwrite=False):
+def snoop(output=None, watch=(), watch_explode=(), depth=1,
+          prefix='', overwrite=False):
     '''
     Snoop on the function, writing everything it's doing to stderr.
 
@@ -54,9 +55,9 @@ def snoop(output=None, variables=(), exploded_variables=(), depth=1, prefix='', 
 
         @pysnooper.snoop('/my/log/file.log')
 
-    See values of some variables that aren't local variables::
+    See values of some expressions that aren't local variables::
 
-        @pysnooper.snoop(variables=('foo.bar', 'self.whatever'))
+        @pysnooper.snoop(watch=('foo.bar', 'self.x["whatever"]'))
 
     Expand values to see all their attributes or items of lists/dictionaries:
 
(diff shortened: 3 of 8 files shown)


cool-RR:
How is this better than `variables=d.keys()`?

alexmojaki:
This shows values, not just keys, so I'm assuming you mean `d.items()`, which really is just as good as `d`. This ensures that all items and their changes are shown and nicely formatted. Otherwise you get the problem in #73, and even when modifications are reported, the user has to manually search through entries to find the differences.

cool-RR:
Ah, I understand. 

I don't like all these classes that people will need to know. I'm sorry, I'm afraid of adding more complexity to the project for features that I don't know whether people will find useful.

alexmojaki:
Also `reprlib` doesn't have good reprs for mappings other than raw `dict`s, so for example in this case:

```python
from collections import OrderedDict

import pysnooper
from pysnooper.variables import Keys

@pysnooper.snoop(variables=(
        Keys('d'),
))
def foo():
    d = OrderedDict()
    d[1] = list(range(20))
    d[2] = list(range(40))

foo()
```

```
20:57:49.925572 call        11 def foo():
20:57:49.925890 line        12     d = OrderedDict()
New var:....... d = OrderedDict()
20:57:49.926047 line        13     d[1] = list(range(20))
New var:....... (d)[1] = [0, 1, 2, 3, 4, 5, ...]
Modified var:.. d = OrderedDict([(1, [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19])])
20:57:49.926290 line        14     d[2] = list(range(40))
New var:....... (d)[2] = [0, 1, 2, 3, 4, 5, ...]
Modified var:.. d = OrderedDict([(1, [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, ...8, 29, 30, 31, 32, 33, 34, 35, 36, 37, 38, 39])])
20:57:49.926611 return      14     d[2] = list(range(40))
Return value:.. None
```

you can really only see what happened in the `(d)[2] =` line, not the `Modified var:.. d = ` line.

alexmojaki:
OK, I added automatic dispatching. It could alternatively use the Mapping and Sequence ABCs, each method has pros and cons. Then I'd suggest documenting the classes under an 'Advanced usage' section for people interested in greater control (making sure the right class is used, excluding keys, slicing indices, and maybe other functions in the future such as depth).

alexmojaki:
I decided to go with the collections ABCs since that's less likely to trigger unwanted behaviour.

Here's a sample of the automatic API:

```python
from types import SimpleNamespace

import pysnooper


@pysnooper.snoop(exploded_variables='d point lst'.split())
def foo():
    d = {'a': 1, 'b': 2, 'c': 'ignore'}
    point = SimpleNamespace(x=3, y=4)
    lst = list(range(5))


foo()
```

```
09:11:50.682783 call         7 def foo():
09:11:50.683004 line         8     d = {'a': 1, 'b': 2, 'c': 'ignore'}
New var:....... (d)['a'] = 1
New var:....... (d)['b'] = 2
New var:....... (d)['c'] = 'ignore'
New var:....... d = {'a': 1, 'b': 2, 'c': 'ignore'}
09:11:50.683432 line         9     point = SimpleNamespace(x=3, y=4)
New var:....... (point).x = 3
New var:....... (point).y = 4
New var:....... point = namespace(x=3, y=4)
09:11:50.683996 line        10     lst = list(range(5))
New var:....... (lst)[0] = 0
New var:....... (lst)[1] = 1
New var:....... (lst)[2] = 2
New var:....... (lst)[3] = 3
New var:....... (lst)[4] = 4
New var:....... lst = [0, 1, 2, 3, 4]
09:11:50.684502 return      10     lst = list(range(5))
Return value:.. None
```

cool-RR:
Could you rebase or resolve conflicts before I review?

cool-RR:
Looks like there are still conflicts to resolve, do you need help with that?

alexmojaki:
I'm not seeing any conflicts.

cool-RR:
See this screenshot: https://www.dropbox.com/s/nd6pthuhr5ubol5/screenshot_HOPPER_2019-04-30_172302.jpg?dl=0

If you'll want help with the conflicts, let me know.

cool-RR:
See this screenshot: https://www.dropbox.com/s/nd6pthuhr5ubol5/screenshot_HOPPER_2019-04-30_172302.jpg?dl=0

If you'll want help with the conflicts, let me know.

alexmojaki:
screen

alexmojaki:
Send me the link to command line instructions and I'll take a look later.

cool-RR:
@alexmojaki Can you please add me as a collaborator here? https://github.com/alexmojaki/PySnooper/

cool-RR:
Sorry for the delay here. Looking good. I rebased and touched up a few places, the result is in the `development` branch. After you give me permission to your repo, I'll push it into this PR.

Here are more things needed to merge:

1. Tests
2. Update the docstring and the readme (or advanced usage, that's good too.)
4. Protect from infinite sequence, and test.
5. Use `ABC`, `@abstractmethod` and such on the classes in variables
6. `_safe_keys` should be a top-level function since it doesn't use `self`
7. No need for `()` in `raise NotImplementedError()`
8. I think you can use vars instead of __dict__ and __slots_. 

I'm willing to split a few items with you if you'd like. You should do at least the first two though, because problems with your design might come up. Let me know which of the items you'd like me to do.

alexmojaki:
3. I'm not iterating blindly anywhere any more as I removed `Enumerate`. `Indices` uses `len` so nothing infinite can happen.
4. The only candidate for `@abstractmethod` is `BaseVariable._items`. Adding it would probably waste reader's time as they think about why it hasn't been used on other methods.
5. `_safe_keys` uses `self`.
7. `vars` simply returns `__dict__`. There was a discussion to make it account for `__slots__` but it was rejected, which disappointed me.

cool-RR:
I agree on 3 and 5.

Regarding 4: I think that `@abstractmethod` is fitting for `_items`.

Regarding 7: Okay, but there should be a test that this logic doesn't fail for a few objects that don't have `__dict__`.

cool-RR:
Looking good. As far as I'm concerned, this only needs the "advanced usage" text file, with a link from the readme, and it's good to merge. Let me know if you'll want help on that.

alexmojaki:
I think we should probably go for a shorter parameter name, like `explode` or `expand`.

cool-RR:
I do think that the argument name is too long, and `explode` would be cool, the problem is that it sounds too much like a boolean and I don't want people to try to pass `True` in. By the way, can you rename from `exploded_variables` to `exploding_variables`? I think the latter sounds cooler.

alexmojaki:
I can attest that it's hard to remember if it's `exploded_variables` or `exploding_variables`. You wrote `exploding` in the issue, but I wrote `exploded` in the code, and then screwed up again and wrote `exploding` in the docs.

cool-RR:
(If you'd like me to do the renaming, I will.)

cool-RR:
Actually, I'll just do it now.

alexmojaki:
It also has nothing to do with variables specifically. It's for arbitrary expressions.

alexmojaki:
On that note, the name `variables` doesn't make sense either, because all variables are tracked, it only makes sense to pass a more complicated expression.

cool-RR:
> On that note, the name `variables` doesn't make sense either, because all variables are tracked, it only makes sense to pass a more complicated expression. I think `watch` would be better to fit with traditional debugger terminology. Maybe `expressions`.

Agreed.

alexmojaki:
Awesome!

What do you think of just `exploding`? Shorter, doesn't sound like a boolean. Bit weird though.

cool-RR:
Sorry, I've grown to really like `watch_explode` :)

alexmojaki:
Oh, I didn't see that one. Commits are moving too fast. That's pretty good.

cool-RR:
Whew, this PR exploded :) Great work Alex, this is a very nice addition to our functionality.

Issue #39, opened 2019-04-24
Feature request: Test `pip install`


cool-RR:
@bittner , this is right up your alley so if you'll feel like doing it, it'll be cool.

It'll be nice if after every release, or after every push to master, there will be separate tests that do `pip install pysnooper` on all platforms and run the tests. This is to rule out any errors in the installation.

bittner:
Every Python environment created by Tox gets `pysnooper` installed. The workflow is roughly:

1. Create virtualenv for Python x.y (e.g. Python 3.7 for `py37`)
1. Run `python setup.py install` (i.e. install `pysnooper`)
1. Install the dependencies (from `deps = ...` in the Tox environment)
1. Run `pytest` (i.e. from `commands = ...`)

If I'm not completely mistaken `pip install pysnooper` does nothing else than download the package from PyPI and run `python setup.py install`. Hence, in theory installing should succeed when Tox runs successfully. And this is now ensured.

Any doubt about this?

bittner:
What is missing indeed is that tests are run on Windows and macOS. We've done this with PythonTurtle already, it's tedious but it's feasible, it's not magic.

And more tests, e.g. integration tests, are always a good idea. Sure!

cool-RR:
Are you sure tox does `setup.py install`? Where is it written?

bittner:
You may want to take a look at the Tox [System Overview](https://tox.readthedocs.io/en/latest/index.html#system-overview) and the related explanations in the Tox documentation.

cool-RR:
Okay, thanks. I'd still welcome a direct test of `pip install pysnooper`. Low priority.

enadol:
On Ubuntu/Raspbian, pip install pysnooper shoots the following:

pi@raspberrypi:~ $ pip3 install pysnooper
Downloading/unpacking pysnooper
  Downloading PySnooper-0.0.22-py2.py3-none-any.whl
Installing collected packages: pysnooper
Cleaning up...
Exception:
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/pip/basecommand.py", line 122, in main
    status = self.run(options, args)
  File "/usr/lib/python3/dist-packages/pip/commands/install.py", line 295, in run
    requirement_set.install(install_options, global_options, root=options.root_path)
  File "/usr/lib/python3/dist-packages/pip/req.py", line 1436, in install
    requirement.install(install_options, global_options, *args, **kwargs)
[...]

PR #76, opened 2019-04-29
Add integration test for installing pysnooper

awarecheetah:
#39: Add tox environment for installing pysnooper.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,14 @@ jobs:
   - { stage: lint, python: 3.7, env: TOXENV=pylint }
   - { stage: lint, python: 3.7, env: TOXENV=bandit }
   - { stage: lint, python: 3.7, env: TOXENV=readme }
+  - { stage: test, python: 3.7, env: TOKENV=pysnooper }
+  - { stage: test, python: 3.6, env: TOKENV=pysnooper }
+  - { stage: test, python: 3.5, env: TOKENV=pysnooper }
+  - { stage: test, python: 3.4, env: TOKENV=pysnooper }
+  - { stage: test, python: 2.7, env: TOKENV=pysnooper }
+  - { stage: test, python: 3.8-ev, env: TOKENV=pysnooper }
+  - { stage: test, python: pypy2.7-6.0, env: TOKENV=pysnooper }
+  - { stage: test, python: pypy3.5, env: TOKENV=pysnooper }
 
   - stage: deploy
     install: skip
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ jobs:
   - { stage: test, python: 3.5, env: TOKENV=pysnooper }
   - { stage: test, python: 3.4, env: TOKENV=pysnooper }
   - { stage: test, python: 2.7, env: TOKENV=pysnooper }
-  - { stage: test, python: 3.8-ev, env: TOKENV=pysnooper }
+  - { stage: test, python: 3.8-dev, env: TOKENV=pysnooper }
   - { stage: test, python: pypy2.7-6.0, env: TOKENV=pysnooper }
   - { stage: test, python: pypy3.5, env: TOKENV=pysnooper }
 
--- a/tox.ini
+++ b/tox.ini
@@ -6,6 +6,7 @@ envlist =
     flake8
     pylint
     bandit
+    pysnooper
     py{27,34,35,36,37,38,py,py3}
     readme
     requirements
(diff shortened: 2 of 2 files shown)


cool-RR:
Cool, can you please add a line that at least imports pysnooper to ensure that it's installed?

awarecheetah:
> Cool, can you please add a line that at least imports pysnooper to ensure that it's installed?

Yes, I will do that with `pip show pysnooper`, which will verify the package name and version.

alexmojaki:
You need a line with `TOXENV=pysnooper` in travis.yml. Don't know what the right stage would be, I guess test?

alexmojaki:
I don't know why but the logs don't show the desired commands being run: https://travis-ci.org/cool-RR/PySnooper/jobs/526204104

tox.ini
```diff
@@ -63,3 +72,8 @@ exclude = .tox,build,dist,pysnooper.egg-info
 
 [pytest]
 addopts = --strict
+
+[pysnooper]
+commands =
+    pip install pysnooper
+    pip show pysnooper
```
bittner: This is unneeded. You can remove it. (Explanation: The `[<tool_name>]` sections are configuration file sections that the tool reads. I don't think pysnooper reads anything out from `tox.ini`.)

tox.ini
```diff
@@ -54,6 +56,13 @@ deps = pip-tools
 commands = pip-compile --output-file requirements.txt requirements.in
 changedir = {toxinidir}
 
+[testenv:pysnooper]
+description = Ensure pip installs pysnooper
+deps = pip
```
bittner: No need to install `pip`. Replace this by `pysnooper` if you want to install it off PyPI (that's probably _not_ what you want), or leave it empty.

tox.ini
```diff
@@ -54,6 +56,13 @@ deps = pip-tools
...
+commands =
+    pip install pysnooper
+    pip show pysnooper
```
bittner: This is not an integration test.

Think: What does this verify? Is the pysnooper code run with it? (It's not.)

Also, installing pysnooper is unneeded, and even if it were you can move it to `deps = `.

tox.ini
```diff
@@ -16,6 +17,7 @@ description = Unit tests
 deps =
     pytest
     python_toolbox
+    pysnooper
```
bittner: This installs the latest released version of pysnooper off PyPI. This is most probably _not_ what you want.

Note that pysnooper (off the checked out commit) is already installed by Tox automatically. Try running `pysnooper` or `python -c "import pysnooper; print(pysnooper.__version__)"` to prove it.

.travis.yml
```diff
@@ -33,6 +33,14 @@ jobs:
...
   - { stage: lint, python: 3.7, env: TOXENV=bandit }
   - { stage: lint, python: 3.7, env: TOXENV=readme }
+  - { stage: test, python: 3.7, env: TOKENV=pysnooper }
+  - { stage: test, python: 3.6, env: TOKENV=pysnooper }
+  - { stage: test, python: 3.5, env: TOKENV=pysnooper }
+  - { stage: test, python: 3.4, env: TOKENV=pysnooper }
+  - { stage: test, python: 2.7, env: TOKENV=pysnooper }
+  - { stage: test, python: 3.8-dev, env: TOKENV=pysnooper }
+  - { stage: test, python: pypy2.7-6.0, env: TOKENV=pysnooper }
+  - { stage: test, python: pypy3.5, env: TOKENV=pysnooper }
```
bittner: This looks very verbose.

What if you simply would run your integration test with the `[testenv]` section? (Then all would not be needed. Hmm?)

bittner:
I think, this contribution can be much simpler:

Simply add an additional command that performs the desired integration test to the `[testenv]`'s [commands](https://github.com/cool-RR/PySnooper/blob/master/tox.ini#L19) attribute, e.g.

```ini
commands =
    pytest
    python -c "print('run my integration test')"
```

I'd also suggest to add `--verbose` [as a pytest switch](https://github.com/cool-RR/PySnooper/blob/master/tox.ini#L64-L65) (as long as you don't have hundreds of tests), e.g.
```ini
[pytest]
addopts = --strict --verbose
```
This makes the output easier to read. I see [you removed it](https://github.com/cool-RR/PySnooper/commits/master/tox.ini), Ram, due to problems with the IDE, hmm.

bittner:
Please note, everyone, that pysnooper is _always_ installed by Tox automatically. That's the whole purpose of a Tox setup: It installs the package under test (using `setup.py`, which is always done unless [skipsdist = true](https://tox.readthedocs.io/en/latest/config.html#conf-skipsdist)).

Take a look at [older builds](https://travis-ci.org/cool-RR/PySnooper/jobs/524525195#L197-L198) to prove yourself. Note the zip (created by setuptools) and the installed pysnooper version (even though it's not specified anyhere explicitly for installation).

awarecheetah:
Thanks for your input @bittner.  I'll make these changes as soon as I can.

awarecheetah:
Actually, I'm just going to close out this PR.

Issue #80, opened 2019-05-02
allow arbitrary callables, perhaps io / file objects, Python loggers for writers


zzzeek:
Looking at https://github.com/cool-RR/PySnooper/blob/master/pysnooper/pysnooper.py#L13 I see our choices for "writier" are, stdout, path to a filename and something called "WritableStream"  .   In our case we have an application that's running under mod_wsgi, so we want to go to a logger.      Since PySnooper's great talent is being able to get deep into a huge stack without any hassle, being able to send the string output anywhere is important.   There are threadsafe loggers as well and things like that which follow built-in conventions like [IO](https://docs.python.org/3/library/io.html), or just a plain Python callable where we can plug anything we want.

zzzeek:
I can PR these things if you are OK with it.

cool-RR:
Allowing a callable is a great idea. Don't forget tests on your PR.

alexmojaki:
To clarify, a `WritableStream` is literally any class with a 'write' method, so it's easy right now to send the output anywhere. But I agree allowing any callable will be convenient, which should be the aim of the library.

zzzeek:
right I just realized that's what the ABC-ness of it does, so that is better.

cool-RR:
Mike, your idea is still good.

PR #81, opened 2019-05-02
Allow a callable to be passed to snoop()

zzzeek:
Fixes: #80

### DIFF
--- a/pysnooper/pysnooper.py
+++ b/pysnooper/pysnooper.py
@@ -23,6 +23,9 @@ def get_write_and_truncate_functions(output):
         def truncate():
             with open(output, 'w') as output_file:
                 pass
+    elif callable(output):
+        write = output
+        truncate = None
     else:
         assert isinstance(output, utils.WritableStream)
         def write(s):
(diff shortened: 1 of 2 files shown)


pysnooper/pysnooper.py
```diff
@@ -23,6 +23,9 @@ def write(s):
         def truncate():
             with open(output, 'w') as output_file:
                 pass
+    elif hasattr(output, "__call__"):
```
alexmojaki:
```suggestion
    elif callable(output):
```

zzzeek: Yes I wasn't sure if you wanted to use callable() here as it was briefly removed in earlier py3k versions but was later restored

cool-RR:
I like this PR, but I don't like the abstractions you added to the test. Please add your test separately without changing the existing test.

Reading material: https://amir.rachum.com/blog/2017/01/14/explicit-tests/

zzzeek:
thank you for your blog post!

cool-RR:
Great job!

PR #85, opened 2019-05-03
Add assert_snoop to refactor tests

alexmojaki:
Based on #81 I'm guessing you're not going to like this. But I read the blog post and I don't think the arguments apply here. The logic that's being duplicated in tests is long and complex, and if in the future that pattern has to be changed in many tests it'll be a pain. Adding new tests is also annoying. This also doesn't repeat any logic from the implementation.

### DIFF
--- a/tests/test_pysnooper.py
+++ b/tests/test_pysnooper.py
@@ -3,13 +3,12 @@
 
 import io
 import textwrap
-import types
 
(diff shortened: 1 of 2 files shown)


cool-RR:
I indeed disagree with this abstraction, sorry.

Issue #79, opened 2019-05-02
add support for thread identifiers


zzzeek:
pysnooper is amazing for the case where we are trying to see where threads are waiting on things.    If I can get the thread identifier in the output as an option, that makes this great.  Here's a hack that makes it work at the moment:

```diff
diff --git a/pysnooper/tracer.py b/pysnooper/tracer.py
index c4f4407..a5768f2 100644
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -4,6 +4,7 @@
 import sys
 import re
 import collections
+import threading
 import datetime as datetime_module
 import itertools
 try:
@@ -226,6 +227,7 @@ class Tracer:
         ### Finished newish and modified variables. ###########################
 
         now_string = datetime_module.datetime.now().time().isoformat()
+        tid_string = threading.get_ident()
         line_no = frame.f_lineno
         source_line = get_source_from_frame(frame)[line_no - 1]
 
@@ -251,7 +253,7 @@ class Tracer:
         #                                                                     #
         ### Finished dealing with misplaced function definition. ##############
 
-        self.write('{indent}{now_string} {event:9} '
+        self.write('{indent}{now_string} {tid_string} {event:9} '
                    '{line_no:4} {source_line}'.format(**locals()))
 
         if event == 'return':

```

cool-RR:
Sounds good. If you're interested, write a PR with a Boolean argument to
snoop, and tests. Ensure you've got padding on the thread ID to keep the
columns good.

alexmojaki:
Perhaps it would be better to allow passing a callable (e.g. `threading.ident`) to `prefix`. More flexible, doesn't increase the number of parameters.

cool-RR:
It's a bit hacky in my opinion. Also they'd have to write the padding logic
themselves.

alexmojaki:
> Also they'd have to write the padding logic themselves

Why? Just use `prefix = self.prefix().strip() + ' '`.

cool-RR:
The point with padding is to have a consistent width even if the length of
the data changes, to keep the columns lined up.

alexmojaki:
Sorry, wasn't thinking straight.

`get_ident()` just returns a number, which is not very useful to humans. We'd probably want to include the name of the current thread too.

alexmojaki:
This can be closed.

PR #87, opened 2019-05-03
Add support for thread identifiers

4383:
Like @zzzeek as suggested these changes display thread infos (identifier and name) on output to help
user to track execution on apps who use threading.

Solve #79

### DIFF
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -7,6 +7,7 @@ import re
 import collections
 import datetime as datetime_module
 import itertools
+import threading
 
 from .variables import CommonVariable, Exploding, BaseVariable
 from .third_party import six, decorator
@@ -128,6 +129,7 @@ class Tracer:
             depth=1,
             prefix='',
             overwrite=False,
+            thread_info=False,
     ):
         '''
         Snoop on the function, writing everything it's doing to stderr.
@@ -162,6 +164,9 @@ class Tracer:
 
             @pysnooper.snoop(prefix='ZZZ ')
 
+        On multi-threaded apps identify which thread are snooped in output::
+
+            @pysnooper.snoop(thread_info=True)
         '''
         self._write, self.truncate = get_write_and_truncate_functions(output)
 
@@ -182,6 +187,8 @@ class Tracer:
         self.prefix = prefix
         self.overwrite = overwrite
         self._did_overwrite = False
+        self.thread_info = thread_info
+        self.thread_info_padding = 0
         assert self.depth >= 1
         self.target_codes = set()
         self.target_frames = set()
@@ -218,6 +225,13 @@ class Tracer:
 
     def _is_internal_frame(self, frame):
         return frame.f_code.co_filename == Tracer.__enter__.__code__.co_filename
+
+    def set_thread_info_padding(self, thread_info):
+        current_thread_len = len(thread_info)
+        self.thread_info_padding = max(self.thread_info_padding,
+                                       current_thread_len)
+        return thread_info.ljust(self.thread_info_padding)
+
 
     def trace(self, frame, event, arg):
 
@@ -281,6 +295,12 @@ class Tracer:
         now_string = datetime_module.datetime.now().time().isoformat()
         line_no = frame.f_lineno
         source_line = get_source_from_frame(frame)[line_no - 1]
+        thread_info = ""
+        if self.thread_info:
+            current_thread = threading.current_thread()
+            thread_info = "{ident}-{name} ".format(
+                ident=current_thread.ident, name=current_thread.getName())
+        thread_info = self.set_thread_info_padding(thread_info)
 
         ### Dealing with misplaced function definition: #######################
         #                                                                     #
@@ -304,7 +324,7 @@ class Tracer:
         #                                                                     #
         ### Finished dealing with misplaced function definition. ##############
 
-        self.write(u'{indent}{now_string} {event:9} '
+        self.write(u'{indent}{now_string} {thread_info}{event:9} '
                    u'{line_no:4} {source_line}'.format(**locals()))
 
         if event == 'return':
--- a/tests/test_pysnooper.py
+++ b/tests/test_pysnooper.py
@@ -3,6 +3,7 @@
 
 import io
 import textwrap
+import threading
 import types
 
 from python_toolbox import sys_tools, temp_file_tools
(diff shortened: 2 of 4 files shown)


4383:
I notice the optional part that you need, I'll update that ASAP on this PR

pysnooper/tracer.py
```diff
@@ -216,7 +220,7 @@ def trace(self, frame, event, arg):
         #                                                                     #
         ### Finished dealing with misplaced function definition. ##############
 
-        self.write(u'{indent}{now_string} {event:9} '
```
alexmojaki: The `u` prefix is important, don't remove it.

cool-RR:
Good job. I'll be waiting for the optional part, and I also want:

1. A test.
2. change `tinfo_string` to `thread_info`.
3. Padding for a consistent column width.
4. Add a line in the readme and docstrings about the option.

alexmojaki:
I still say we should go for a callable prefix. Whatever system we use to keep the column width consistent in this PR, why can't we use it for a callable prefix? Then we could supply and document a utility function `thread_info` or something for this common so that the user doesn't have to write their own, they just use `prefix=thread_info`.

cool-RR:
I understand, but I still think it's too hacky.

4383:
I need to solve the padding issue

zzzeek:
How do you want to apply padding if we don't know the names of all thread identifiers (including those that might not exist yet, so even diving into threading.enumerate() not good enough) and therefore how much length is needed?

4383:
The main issue come from the name of the thread to padding output. 
We can't know by advance all threads names... I think using only thread ID is enough.

cool-RR:
Guess a length that'll accommodate most thread names, and for the few that
are longer, truncate them.

alexmojaki:
I say keep track of the padding so far and increase it as necessary, never decreasing it. A couple of bumps in the logs isn't a big problem. Use enumerate to make a good guess to start with, maybe add a couple of extra spaces initially to reduce the likelihood of bumps. Truncation risks completely misleading the user, e.g. if `Thread-11` is truncated to `Thread-1`.

pysnooper/tracer.py
```diff
@@ -216,8 +222,14 @@ def trace(self, frame, event, arg):
...
-                   u'{line_no:4} {source_line}'.format(**locals()))
+        if self.threads_info:
+            output = u'{indent}{now_string} {thread_info} {event:9} ' \
+                     u'{line_no:4} {source_line}'.format(**locals())
```
alexmojaki: 1. Don't duplicate this logic.
2. Don't needlessly compute `thread_info`.

```python
if self.threads_info:
    current_thread = threading.current_thread()
    thread_info = "{ident}-{name} ".format(
        ident=current_thread.ident, name=current_thread.getName())
else:
    thread_info = ""
    
output = u'{indent}{now_string} {thread_info}{event:9} ' \
         u'{line_no:4} {source_line}'.format(**locals())
```

tests/test_pysnooper.py
```diff
@@ -41,6 +41,35 @@ def my_function(foo):
     )
 
 
+def test_threads_info():
+    string_io = io.StringIO()
```
alexmojaki: This is unused. A good example of where #85 would help.

tests/utils.py
```diff
@@ -165,17 +165,27 @@ def _check_value(self, value):
...
+                r"""(?P<threads_info>[0-9]*-[A-Za-z_]*) """
+                r"""(?P<event_name>[a-z_]*) +(?P<line_number>[0-9]*) """
+                r"""+(?P<source>.*)$""" % (re.escape(self.prefix,))
+            )
+        else:
+            self.line_pattern = re.compile(
+                r"""^%s(?P<indent>(?: {4})*)[0-9:.]{15} """
+                r"""(?P<event_name>[a-z_]*) +(?P<line_number>[0-9]*) """
+                r"""+(?P<source>.*)$""" % (re.escape(self.prefix,))
+            )
```
alexmojaki: Stick to a single regex and make the group optional (`(...)?`).

tests/utils.py
```diff
@@ -197,7 +207,10 @@ def check(self, s):
         match = self.line_pattern.match(s)
         if not match:
             return False
-        _, event_name, _, source = match.groups()
+        if self.threads_info:
+            _, _, event_name, _, source = match.groups()
+        else:
+            _, event_name, _, source = match.groups()
```
alexmojaki: Continuing from the above, `groups` will have `None` for the optional group if it isn't present. Although at this point using `groupdict` and the group names might be better.

cool-RR:
> I say keep track of the padding so far and increase it as necessary, never decreasing it. 

Agreed. I also agree about the groupdict

4383:
Changes asked added, and padding added now.

Please take a look to this [POC](https://gist.github.com/4383/6b8160c1db5dae411ec91ba505d27cfe) to see sample output with padding.

pysnooper/tracer.py
```diff
@@ -133,6 +137,16 @@ def __enter__(self):
...
         sys.settrace(self.original_trace_function)
 
+    def set_padding(self, thread_info):
+        current_thread_len = len(thread_info)
+        if self.padding < current_thread_len:
+            self.padding = current_thread_len
+            return thread_info
+        else:
+            needed_spaces = " " * (self.padding - current_thread_len)
+            return "{thread_info}{needed_spaces}".format(**locals())
```
alexmojaki: Try:

```python
self.padding = max(self.padding, current_thread_len)
return thread_info.ljust(self.padding)
```

pysnooper/tracer.py
```diff
@@ -117,6 +119,8 @@ def __init__(self, target_code_object, write, truncate, watch=(),
         self.prefix = prefix
         self.overwrite = overwrite
         self._did_overwrite = False
+        self.threads_info = threads_info
+        self.padding = 0
```
alexmojaki: I think `padding` and `set_padding` sound too generic. They should indicate that they specifically apply to threads.

pysnooper/tracer.py
```diff
@@ -193,6 +207,12 @@ def trace(self, frame, event, arg):
         now_string = datetime_module.datetime.now().time().isoformat()
         line_no = frame.f_lineno
         source_line = get_source_from_frame(frame)[line_no - 1]
+        thread_info = ""
+        if self.threads_info:
+            current_thread = threading.current_thread()
+            thread_info = "{ident}-{name} ".format(
+                ident=current_thread.ident, name=current_thread.getName())
+        thread_info = self.set_padding(thread_info)
```
alexmojaki: Why isn't this also under the if?

4383: To keep padding with snooped where thread info is disable.
IHMO I guess we need too.

4383: look at the POC and sample output

tests/utils.py
```diff
@@ -173,6 +173,7 @@ def __init__(self, source=None, source_regex=None, prefix=''):
             assert source_regex is None
         self.line_pattern = re.compile(
             r"""^%s(?P<indent>(?: {4})*)[0-9:.]{15} """
+            r"""(?P<threads_info>[0-9]*-[A-Za-z_]* )?"""
```
alexmojaki: 1. Should be `+`, not `*`. Should probably be the same for the rest of the regex as well, don't know what the reasoning was.
2. Need to allow for more than one space.
3. Need to allow for `-` and numbers in the thread name. For me, `threading.Thread().name` gives `'Thread-357'`.

You also need to write a test that has multiple threads with names of different lengths.

pysnooper/tracer.py
4383: See me next comment on padding on everything by default to keep the alignment between snooped with threads info turned on and turned off in the same output.

pysnooper/tracer.py
alexmojaki: Once `Tracer.threads_info` is true, it's not going to become false unless the user does something quite weird. But it's not impossible, especially as this library changes, so I guess this can stay.

pysnooper/tracer.py
alexmojaki: I'm just talking about the names. I would go for `thread_info_padding` and `set_thread_info_padding`

alexmojaki:
@cool-RR should the parameter name be `thread_info` or `threads_info`?

pysnooper/tracer.py
cool-RR: @4383 I agree with Alex's comments, `ljust` is better.

tests/utils.py
cool-RR: Agreed, except I'm not sure about 3. But it's a small point anyway, so I don't mind about that.

cool-RR:
> @cool-RR should the parameter name be `thread_info` or `threads_info`? I'm leaning towards singular.

I see no reason why it should be `threads_info`. You're keeping information about only one thread in it, right? If so, then it's singular.

cool-RR:
@4383 You're not waiting for us right? Just making sure.

alexmojaki:
I'm sorry but I would actually avoid spending more effort on this because my PR is going to completely obsolete it and I have something else more generalised line up afterwards.

cool-RR:
Can you add a bit of detail?

cool-RR:
@4348 Separately, I added you now to the AUTHORS file because regardless of whether this PR makes it through, you've made a contribution.

cool-RR:
> I need to add the test with multi threads but I'm not convinced by the benefits to test that, I guess a better approach can be to add unit test on the padding methods and mechanism to avoid errors on it instead of testing something more bluring and general. We already test for output with threads turned on so using one or more threads doesn't change the deal...

A test with multiple threads are essential. This test should also test for the padding increase. Regarding Alex's separate refactoring that might obsolete this PR: I can't know when I'll have time to review that, so it shouldn't block this PR.

alexmojaki:
> Can you add a bit of detail?

I've got code which lets the user specify which columns they want in the output and in what order. It handles padding for all of them.

```python
from pysnooper import snoop

@snoop(columns='time thread thread_ident file function', depth=2)
def foo(x):
    y = x + 1
    y += barlo(y)
    z = list(range(1000))
    return z

def barlo(z):
    return z * 2

foo(4)
```

```
Starting var:................................................ x = 4
08:04:22.399994 MainThread 140736070026112 scratch_220.py foo call         5 def foo(x):
08:04:22.429134 MainThread 140736070026112 scratch_220.py foo line         6     y = x + 1
New var:..................................................... y = 5
08:04:22.465483 MainThread 140736070026112 scratch_220.py foo line         7     y += barlo(y)
    Starting var:.................................................. z = 5
    08:04:22.503480 MainThread 140736070026112 scratch_220.py barlo call        12 def barlo(z):
    08:04:22.533012 MainThread 140736070026112 scratch_220.py barlo line        13     return z * 2
    08:04:22.563545 MainThread 140736070026112 scratch_220.py barlo return      13     return z * 2
    Return value:.................................................. 10
Modified var:.................................................. y = 15
08:04:22.610026 MainThread 140736070026112 scratch_220.py foo   line         8     z = list(range(1000))
New var:....................................................... z = [0, 1, 2, 3, 4, 5, ...]
08:04:22.654862 MainThread 140736070026112 scratch_220.py foo   line         9     return z
08:04:22.687353 MainThread 140736070026112 scratch_220.py foo   return       9     return z
Return value:.................................................. [0, 1, 2, 3, 4, 5, ...]
```

The user can also pass callables to make custom columns.

Even without this, if the formatting PR goes through, then the thread info stuff has to be in the formatter, and the test will have to like `test_sample` rather than asserting entries.

cool-RR:
Okay, interesting. I need to think about it.

4383:
> I'm sorry but I would actually avoid spending more effort on this because my PR is going to completely obsolete it and I have something else more generalised line up afterwards.

@alexmojaki @cool-RR you still want these changes? Not sure to understand.

I don't want to spend my time on something who will finish to the trash...

alexmojaki:
Like I said, I think you should wait for now. We may come back to it later.

cool-RR:
@alexmojaki Okay, let's wait with this. @4383 : Sorry for the frustration!

4383:
Well, finally I've added the missing test (multi thread and padding test). 

If you don't request new changes, I think everything is done on my side...

cool-RR:
@4383 Since I decided against the big refactoring, I'm interested in this PR. I think it's almost ready, except one thing: Your tests don't really test that the thread reported is the right thread. That's kind of... The main thing isn't it? 

Modify `_BaseEventEntry.__init__` to accept `thread_info` and `thread_info_regex` similarly to `source` and `source_regex`, and then in your tests add these arguments to ensure that the correct thread is being reported.

alexmojaki:
If the variable lines are also supposed to be aligned, e.g:

```
Starting var:............................. x = 4
08:04:22.399994 MainThread-140736070026112 call         5 def foo(x):
```

then that has to be accounted for in the implementation as well.

cool-RR:
> If the variable lines are also supposed to be aligned, e.g:

I don't think that's important enough to implement.

4383:
> @4383 Since I decided against the big refactoring, I'm interested in this PR. I think it's almost ready, except one thing: Your tests don't really test that the thread reported is the right thread. That's kind of... The main thing isn't it?
> 
> Modify `_BaseEventEntry.__init__` to accept `thread_info` and `thread_info_regex` similarly to `source` and `source_regex`, and then in your tests add these arguments to ensure that the correct thread is being reported.

@cool-RR thanks for your feedback ack!
I'll submit changes related to your comments

4383:
Oh wait I need to do some changes

4383:
done!

cool-RR:
Great job! I'm happy you persevered through this long PR process!

4383:
> Great job! I'm happy you persevered through this long PR process!

@cool-RR My pleasure, you are welcome! :)

PR #90, opened 2019-05-03
add usage to package docstring; fixes #83

### LINKED ISSUE #83: Suggestion: "pydoc pysnooper" should show how to use this

andreasvc:
Aside, does "pysnooper" need to be a package? If it were a module, "pydoc pysnooper" would directly show the docstring of "pysnooper.snoop"; now there's an extra level of indirection.

### DIFF
--- a/pysnooper/__init__.py
+++ b/pysnooper/__init__.py
@@ -1,5 +1,20 @@
 # Copyright 2019 Ram Rachum and collaborators.
 # This program is distributed under the MIT license.
+"""PySnooper - Never use print for debugging again
+
+Usage:
+
+    import pysnooper
+
+    @pysnooper.snoop()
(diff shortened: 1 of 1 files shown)


alexmojaki:
Alternatively you can just put `__doc__ = snoop.__doc__` at the end of the module.

andreasvc:
You can put any string there, but that wouldn't show the function
signature properly. Have a look at "pydoc re" to see how a module and
its classes and functions can be documented properly without having to
jump to other places.

alexmojaki:
I mostly meant doing that to not have to put any documentation in the module at all. In answer to your other concern, I think it's always going to be a package unless you can make it a module while still keeping all the other files separate.

cool-RR:
Thanks @andreasvc !

PR #96, opened 2019-05-04
Allow using snoop as context manager

alexmojaki:
This lets the library be used like so:

```python
import pysnooper

def foo():
    for i in range(100):
        pass
    with pysnooper.snoop():
        y = 2

foo()
```

Output:

```
New var:....... i = 99
12:28:08.659157 line         8         y = 2
```

This way the user can easily select parts of the code to trace without having to extract it into a function. It can even be used outside of a function.

I've also done some refactoring which I think simplifies the codebase overall (e.g. less indirection, no duplicated arguments) and is good in its own right, although you might disagree. For this feature in particular, it made things easier.


cool-RR:
Good idea. I'll review.

cool-RR:
I like both the feature and the refactoring. I don't like that you removed the `get_write_and_truncate_functions` function. The `__init__` method is crowded enough as it is. 

I'm a little suspicious that this would fail for some cases. Go ahead and write some tests so we could be more confident :)

alexmojaki:
> I like both the feature and the refactoring. I don't like that you removed the get_write_and_truncate_functions function. The __init__ method is crowded enough as it is.

I agree. It fell out in the process and I half expected this whole PR to get rejected anyway.

cool-RR:
Great job.

1. The `_should_trace_frame` refactoring is a problem. Consider that the trace function is executed on every single line. It's important that when the logic isn't needed, it's terminated as soon as possible. (Because that can happen thousands of times.) Therefore the very beginning of the trace function, which weeds out the irrelevant calls, should be highly optimized. This is why I didn't use a function call there.
2. Documentation. This shouldn't be in advanced usage, this should be near the top of the examples.

alexmojaki:
I understand the reasoning for optimising that line, but I also didn't want to duplicate code. I've inlined the function. Note, however, that the trace function is no longer being called for every line - for code we don't want to trace, only `call` events are being generated. So when we get to that line, the interpreter is already doing one function call for some external code, plus a call to `trace`.

I've added some documentation, feel free to make any changes to it you'd like.

alexmojaki:
Please can this be merged ASAP, I'm itching to move forward. See https://github.com/alexmojaki/PySnooper/pull/1.

cool-RR:
Great job! I didn't know about `f_trace`, that's very helpful.

alexmojaki:
Just drawing your attention two the commits you're mixing up:

https://github.com/cool-RR/PySnooper/pull/96/commits/cb1ea3257a49ae7f8b0b2282eef887003fb05f55 and https://github.com/cool-RR/PySnooper/pull/96/commits/62c4dad496964067fcee1cfc94aa458c94d47f5e

cool-RR:
I understand, thanks! I also did an experiment now to see for myself how `settrace` doesn't work on the current frame. I never knew that!

PR #97, opened 2019-05-04
[WIP] Include filenames when depth >= 2

codexetreme:
I have added some lines of code that are supposed to implement this feature

### DIFF
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -212,12 +212,20 @@ class Tracer:
                     # Found the def line!
                     line_no = candidate_line_no
                     source_line = candidate_source_line
-                    file_name = (frame.f_code.co_filename if self.depth>=2 else '')
-                    self.write('In file:.. {file_name}'.format(**locals()))
                     break
         #                                                                     #
         ### Finished dealing with misplaced function definition. ##############
 
+        ### Printing the filename if depth>=2 #################################
+        #                                                                     #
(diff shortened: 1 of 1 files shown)


codexetreme:
So, the main issue I face is, it is printing the file name for the first file, and not for any more subsequent files or for function calls

test_pysnooper_depth.py

```
import os
from ys import prep

import pysnooper
@pysnooper.snoop(depth=2)
def print_dir(path):
	print (os.listdir(path)[:3])
	wassup()
	prep()


def wassup():
	print ('hello')

print_dir('/home/codexetreme/Documents/')
```


ys.py

```
def prep():
    print ('yay')
```

codexetreme:
Working on: #6

cc: @cool-RR

alexmojaki:
Look at the context where you've added the code. It's only for calls to decorated functions.

cool-RR:
@codexetreme See Alex's comments.

codexetreme:
Ok, I have fixed the issue, for the most part, my question now is, should I make changes to the tests (I have created one for testing this) as they seem to be failing since there is an unaccounted line in the output log (the line that shows the file name if depth >=2)

cool-RR:
The tests should be modified to pass, yes. And your entire business logic should start with `if self.depth >= 2` because most users aren't going to use that.

cool-RR:
Closed because of low activity.

PR #102, opened 2019-05-08
Cache source by both module name and filename. Fixes #101

### LINKED ISSUE #101: IndexError: list index out of range in jupyter
first I run 
```@pysnooper.snoop()
def number_to_bits(number):
    if number:
        bits = []
        while number:
            number, remainder = divmod(number, 2)
            bits.insert(0, remainder)
        return bits
    else:
        return [0]

number_to_bits(6)
```
It works correctly.

Then I execute 
```
def foo():
    lst = []
    for i in range(10):

    with pysnooper.snoop():
        lower = min(lst)
        upper = max(lst)
        mid = (lower + upper) / 2
        print(lower, mid, upper)

foo()
```

it turns up some errors:
```
/home/ubuntu/.local/lib/python2.7/site-packages/pysnooper/tracer.pyc in trace(self, frame, event, arg)
    289         now_string = datetime_module.datetime.now().time().isoformat()
    290         line_no = frame.f_lineno
--> 291         source_line = get_source_from_frame(frame)[line_no - 1]
    292 
    293         ### Dealing with misplaced function definition: #######################

IndexError: list index out of range
```

If I only execute the second function foo() after restarting the jupyter kernel, the code works correctly.

alexmojaki:


cool-RR:
Great fix!

PR #107, opened 2019-05-10
Keep proper track of previous tracing functions

alexmojaki:
Consider this script:

```python
import sys
import pysnooper

@pysnooper.snoop()
def factorial(x):
    if x <= 1:
        return 1
    return x * factorial(x - 1)

factorial(4)

print(sys.gettrace())
```

When run against master, at the end it prints `<bound method Tracer.trace of <pysnooper.tracer.Tracer object at 0x10f854cc0>>` because that's what `original_trace_function` became in the process, so the original state was lost. This PR fixes that.

### DIFF
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -7,6 +7,7 @@ import re
 import collections
 import datetime as datetime_module
 import itertools
+import threading
 
 from .variables import CommonVariable, Exploding, BaseVariable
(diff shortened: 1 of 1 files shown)


cool-RR:
Looks good. Did you test it manually on cases where it's relevant?

cool-RR:
Oh, I missed the recursion in your example, and I guess you tested on that.

PR #108, opened 2019-05-10
Indent based on how many calls are currently being traced

alexmojaki:
Instead of:

```
Starting var:.. x = 4
16:33:15.380118 call         7 def factorial(x):
16:33:15.380440 line         8     if x <= 1:
16:33:15.380514 line        10     return x * factorial(x - 1)
Starting var:.. x = 3
16:33:15.380631 call         7 def factorial(x):
16:33:15.380695 line         8     if x <= 1:
16:33:15.380745 line        10     return x * factorial(x - 1)
Starting var:.. x = 2
16:33:15.380840 call         7 def factorial(x):
16:33:15.380896 line         8     if x <= 1:
16:33:15.380944 line        10     return x * factorial(x - 1)
Starting var:.. x = 1
16:33:15.381036 call         7 def factorial(x):
16:33:15.381096 line         8     if x <= 1:
16:33:15.381146 line         9         return 1
16:33:15.381197 return       9         return 1
Return value:.. 1
16:33:15.381285 return      10     return x * factorial(x - 1)
Return value:.. 2
16:33:15.381371 return      10     return x * factorial(x - 1)
Return value:.. 6
16:33:15.381456 return      10     return x * factorial(x - 1)
Return value:.. 24
```

Shows:

```
Starting var:.. x = 4
16:33:02.873205 call         7 def factorial(x):
16:33:02.873451 line         8     if x <= 1:
16:33:02.873515 line        10     return x * factorial(x - 1)
    Starting var:.. x = 3
    16:33:02.873629 call         7 def factorial(x):
    16:33:02.873699 line         8     if x <= 1:
    16:33:02.873756 line        10     return x * factorial(x - 1)
        Starting var:.. x = 2
        16:33:02.873863 call         7 def factorial(x):
        16:33:02.873931 line         8     if x <= 1:
        16:33:02.874051 line        10     return x * factorial(x - 1)
            Starting var:.. x = 1
            16:33:02.874157 call         7 def factorial(x):
            16:33:02.874222 line         8     if x <= 1:
            16:33:02.874276 line         9         return 1
            16:33:02.874330 return       9         return 1
            Return value:.. 1
        16:33:02.874589 return      10     return x * factorial(x - 1)
        Return value:.. 2
    16:33:02.874703 return      10     return x * factorial(x - 1)
    Return value:.. 6
16:33:02.874813 return      10     return x * factorial(x - 1)
Return value:.. 24
```

### DIFF
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -246,6 +246,7 @@ class Tracer:
         # or the user asked to go a few levels deeper and we're within that
         # number of levels deeper.
 
+        depth = 0
         if not (frame.f_code in self.target_codes or frame in self.target_frames):
             if self.depth == 1:
                 # We did the most common and quickest check above, because the
@@ -256,17 +257,19 @@ class Tracer:
                 return None
             else:
                 _frame_candidate = frame
-                for i in range(1, self.depth):
+                for depth in range(1, self.depth):
                     _frame_candidate = _frame_candidate.f_back
                     if _frame_candidate is None:
                         return None
                     elif _frame_candidate.f_code in self.target_codes or _frame_candidate in self.target_frames:
-                        indent = ' ' * 4 * i
                         break
                 else:
                     return None
-        else:
-            indent = ''
+
+        stack = self.thread_local.original_trace_functions
+        depth += len(stack) - 1
+        indent = ' ' * 4 * depth
+
         #                                                                     #
         ### Finished checking whether we should trace this line. ##############
 
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -121,6 +121,9 @@ def get_write_and_truncate_functions(output):
     return write, truncate
 
 
+thread_global = threading.local()
+
+
 class Tracer:
     def __init__(
             self,
@@ -246,7 +249,6 @@ class Tracer:
         # or the user asked to go a few levels deeper and we're within that
         # number of levels deeper.
 
-        depth = 0
         if not (frame.f_code in self.target_codes or frame in self.target_frames):
             if self.depth == 1:
                 # We did the most common and quickest check above, because the
@@ -257,7 +259,7 @@ class Tracer:
                 return None
             else:
                 _frame_candidate = frame
-                for depth in range(1, self.depth):
+                for i in range(1, self.depth):
                     _frame_candidate = _frame_candidate.f_back
                     if _frame_candidate is None:
                         return None
@@ -266,9 +268,10 @@ class Tracer:
                 else:
                     return None
 
-        stack = self.thread_local.original_trace_functions
-        depth += len(stack) - 1
-        indent = ' ' * 4 * depth
+        thread_global.__dict__.setdefault('depth', -1)
+        if event == 'call':
+            thread_global.depth += 1
+        indent = ' ' * 4 * thread_global.depth
 
         #                                                                     #
         ### Finished checking whether we should trace this line. ##############
@@ -339,5 +342,6 @@ class Tracer:
             self.write('{indent}Return value:.. {return_value_repr}'.
                                                             format(**locals()))
             del self.frame_to_local_reprs[frame]
+            thread_global.depth -= 1
 
         return self.trace
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,3 @@
+import pytest
+
+pytest.register_assert_rewrite('tests.utils')
--- a/tests/samples/indentation.py
+++ b/tests/samples/indentation.py
@@ -1,0 +1,43 @@
+import pysnooper
+
+
+@pysnooper.snoop(depth=2)
+def main():
+    f2()
+
+
+def f2():
+    f3()
+
+
+def f3():
+    f4()
+
+
+@pysnooper.snoop(depth=2)
+def f4():
+    f5()
+
+
+def f5():
+    pass
+
+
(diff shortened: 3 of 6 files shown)


cool-RR:
That's nice, but it seems like an edge case to me. Also I don't know how
it's going to interact with the indents used for depth > 1.

alexmojaki:
> it seems like an edge case to me

It's for any case where multiple snooped frames are in the stack, which is very likely if you decorate more than one function. It's not limited to recursion.

> Also I don't know how it's going to interact with the indents used for depth > 1.

I tried to create a demo and ran into #109, I'll revisit this after that gets sorted.

cool-RR:
Okay, if interaction with depths > 1 is proven to be good (tests would be good) then I'll likely accept.

alexmojaki:
Here's an example:

```python
import pysnooper


@pysnooper.snoop(depth=2)
def factorial(x):
    if x <= 1:
        return 1
    return mul(x, factorial(x - 1))


def mul(a, b):
    return a * b


factorial(4)
```

```
Starting var:.. x = 4
20:28:17.875295 call         5 def factorial(x):
20:28:17.875509 line         6     if x <= 1:
20:28:17.875550 line         8     return mul(x, factorial(x - 1))
    Starting var:.. x = 3
    20:28:17.875624 call         5 def factorial(x):
    20:28:17.875668 line         6     if x <= 1:
    20:28:17.875703 line         8     return mul(x, factorial(x - 1))
        Starting var:.. x = 2
        20:28:17.875771 call         5 def factorial(x):
        20:28:17.875813 line         6     if x <= 1:
        20:28:17.875849 line         8     return mul(x, factorial(x - 1))
            Starting var:.. x = 1
            20:28:17.875913 call         5 def factorial(x):
            20:28:17.875953 line         6     if x <= 1:
            20:28:17.875987 line         7         return 1
            20:28:17.876021 return       7         return 1
            Return value:.. 1
            Starting var:.. a = 2
            Starting var:.. b = 1
            20:28:17.876111 call        11 def mul(a, b):
            20:28:17.876151 line        12     return a * b
            20:28:17.876190 return      12     return a * b
            Return value:.. 2
        20:28:17.876235 return       8     return mul(x, factorial(x - 1))
        Return value:.. 2
        Starting var:.. a = 3
        Starting var:.. b = 2
        20:28:17.876320 call        11 def mul(a, b):
        20:28:17.876359 line        12     return a * b
        20:28:17.876397 return      12     return a * b
        Return value:.. 6
    20:28:17.876442 return       8     return mul(x, factorial(x - 1))
    Return value:.. 6
    Starting var:.. a = 4
    Starting var:.. b = 6
    20:28:17.876525 call        11 def mul(a, b):
    20:28:17.876563 line        12     return a * b
    20:28:17.876601 return      12     return a * b
    Return value:.. 24
20:28:17.876646 return       8     return mul(x, factorial(x - 1))
Return value:.. 24
```

alexmojaki:
OK, the whole thing is a little more complicated than I imagined. Here's another example:

```python
import pysnooper

snoop = pysnooper.snoop(depth=2)

@snoop
def f1():
    f2()

def f2():
    f3()

def f3():
    f4()

@snoop
def f4():
    f5()

def f5():
    pass

f1()
```

```
20:33:22.941620 call         7 def f1():
20:33:22.942314 line         8     f2()
    20:33:22.942534 call        11 def f2():
    20:33:22.942693 line        12     f3()
    20:33:22.942820 call        20 def f4():
    20:33:22.942918 line        21     f5()
        20:33:22.943023 call        24 def f5():
        20:33:22.943105 line        25     pass
        20:33:22.943164 return      25     pass
        Return value:.. None
    20:33:22.943243 return      21     f5()
    Return value:.. None
    20:33:22.943344 return      12     f3()
    Return value:.. None
20:33:22.943409 return       8     f2()
Return value:.. None
```

1. You have to reuse the same tracer object to get this indentation, because it's based on tracer state. To allow separate calls to `pysnooper.snoop`, the thread local storage needs to be in a global variable.
2. I would expect an extra indent at the call to f4 (and f5 by implication) but it's not there because it only looks at the `depth` in relation to the most recently traced function, not throughout the stack.

alexmojaki:
OK, it wasn't as difficult or messy as imagined, and both problems could be solved at once. Here's an updated example, no reusing of tracers:

```python
import pysnooper


@pysnooper.snoop(depth=2)
def f1():
    f2()


def f2():
    f3()


def f3():
    f4()


@pysnooper.snoop(depth=2)
def f4():
    f5()


def f5():
    pass


f1()
```

```
21:10:42.298924 call         5 def f1():
21:10:42.299158 line         6     f2()
    21:10:42.299205 call         9 def f2():
    21:10:42.299246 line        10     f3()
        21:10:42.299305 call        18 def f4():
        21:10:42.299348 line        19     f5()
            21:10:42.299386 call        22 def f5():
            21:10:42.299424 line        23     pass
            21:10:42.299460 return      23     pass
            Return value:.. None
        21:10:42.299509 return      19     f5()
        Return value:.. None
    21:10:42.299577 return      10     f3()
    Return value:.. None
21:10:42.299627 return       6     f2()
Return value:.. None
```

cool-RR:
Great work with the thread global! Do you want to make a test out of one of your examples? It's touching on sensitive code so I want to be careful.

cool-RR:
Ready for merge?

alexmojaki:
Yes

PR #110, opened 2019-05-10
Truncate long variables in the middle

alexmojaki:
In general there may be useful information at the end of a variable. This also solves the problem in #73

### DIFF
--- a/pysnooper/utils.py
+++ b/pysnooper/utils.py
@@ -54,10 +54,16 @@ def get_shortish_repr(item):
     except Exception:
         r = 'REPR FAILED'
     r = r.replace('\r', '').replace('\n', '')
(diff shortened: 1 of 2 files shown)


cool-RR:
Good idea. But instead of `string[:left], string[-right:]` you can just do `string[:left], string[left:]` and not calculate `right`, right?

alexmojaki:
They will be different if `(max_length - 3)` is odd. The `-3` is not very important either. It's a perfectionist function. If you prefer I can simplify the function and let it be a few characters off.

cool-RR:
> They will be different if `(max_length - 3)` is odd.

Oops, I tried to be clever and made a basic mistake :)

PR #111, opened 2019-05-10
Delete frames when exiting to free up memory

alexmojaki:
`Tracer` currently holds on to every frame it has traced, including all its local variables (not just their reprs). Try this script:

```python
import pysnooper

@pysnooper.snoop()
def foo():
    lst = list(range(1000))

while 1:
    foo()
```

and check `top`. The memory usage skyrockets.


cool-RR:
Good catch!

PR #113, opened 2019-05-10
Preserve the local variables order in the output.

alexbers:
Write the variables in the declare order, instead of the alphabetical one.

This causes function arguments to be printed in the declare order.
All watched variables are outputted after declared ones.

Example code:

```python
import pysnooper

@pysnooper.snoop()
def f(one, two, three, four):
    five, six, seven = 5, 6, 7

f(1, 2, 3, 4)
```

Before:
```
Starting var:.. four = 4
Starting var:.. one = 1
Starting var:.. three = 3
Starting var:.. two = 2
03:22:33.398756 call         4 def f(one, two, three, four):
03:22:33.398987 line         5     five, six, seven = 5, 6, 7
New var:....... five = 5
New var:....... seven = 7
New var:....... six = 6
03:22:33.399147 return       5     five, six, seven = 5, 6, 7
Return value:.. None
```

After:
```
Starting var:.. one = 1
Starting var:.. two = 2
Starting var:.. three = 3
Starting var:.. four = 4
03:23:22.381525 call         4 def f(one, two, three, four):
03:23:22.381743 line         5     five, six, seven = 5, 6, 7
New var:....... five = 5
New var:....... six = 6
New var:....... seven = 7
03:23:22.381872 return       5     five, six, seven = 5, 6, 7
Return value:.. None
```

### DIFF
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -288,9 +288,7 @@ class Tracer:
             if name not in old_local_reprs:
                 self.write('{indent}{newish_string}{name} = {value_repr}'.format(
                                                                        **locals()))
-
-        for name, value_repr in local_reprs.items():
-            if name in old_local_reprs and old_local_reprs[name] != value_repr:
+            elif old_local_reprs[name] != value_repr:
                 self.write('{indent}Modified var:.. {name} = {value_repr}'.format(
                                                                    **locals()))
 
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -19,17 +19,12 @@ ipython_filename_pattern = re.compile('^<ipython-input-([0-9]+)-.*>$')
 
 
 def get_local_reprs(frame, watch=()):
-    result_items = [(key, utils.get_shortish_repr(value)) for key, value
-                                                     in frame.f_locals.items()]
+    result_items = [(key, utils.get_shortish_repr(value)) for key, value in frame.f_locals.items()]
 
     vars_order = frame.f_code.co_varnames + frame.f_code.co_cellvars + frame.f_code.co_freevars
-    result_items.sort(
-        key=lambda key_value: (
-            (vars_order.index(key_value[0]) if key_value[0] in vars_order else float("inf")),
-            key_value[0]
-        )
-    )
+    vars_order += tuple(frame.f_locals.keys())
 
+    result_items.sort(key=lambda key_value: vars_order.index(key_value[0]))
     result = collections.OrderedDict(result_items)
 
     for variable in watch:
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -193,7 +193,7 @@ class Tracer:
              v if isinstance(v, BaseVariable) else Exploding(v)
              for v in utils.ensure_tuple(watch_explode)
         ]
-        self.frame_to_local_reprs = collections.OrderedDict()
+        self.frame_to_local_reprs = {}
         self.depth = depth
         self.prefix = prefix
         self.overwrite = overwrite
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -19,11 +19,10 @@ ipython_filename_pattern = re.compile('^<ipython-input-([0-9]+)-.*>$')
 
 
 def get_local_reprs(frame, watch=()):
-    result_items = [(key, utils.get_shortish_repr(value)) for key, value in frame.f_locals.items()]
-
     vars_order = frame.f_code.co_varnames + frame.f_code.co_cellvars + frame.f_code.co_freevars
     vars_order += tuple(frame.f_locals.keys())
 
+    result_items = [(key, utils.get_shortish_repr(value)) for key, value in frame.f_locals.items()]
     result_items.sort(key=lambda key_value: vars_order.index(key_value[0]))
     result = collections.OrderedDict(result_items)
 
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -19,7 +19,8 @@ ipython_filename_pattern = re.compile('^<ipython-input-([0-9]+)-.*>$')
 
 
 def get_local_reprs(frame, watch=()):
-    vars_order = frame.f_code.co_varnames + frame.f_code.co_cellvars + frame.f_code.co_freevars + tuple(frame.f_locals.keys())
+    code = frame.f_code
+    vars_order = code.co_varnames + code.co_cellvars + code.co_freevars + tuple(frame.f_locals.keys())
 
     result_items = [(key, utils.get_shortish_repr(value)) for key, value in frame.f_locals.items()]
     result_items.sort(key=lambda key_value: vars_order.index(key_value[0]))
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -282,23 +282,19 @@ class Tracer:
         self.frame_to_local_reprs[frame] = local_reprs = \
                                        get_local_reprs(frame, watch=self.watch)
 
-        modified_local_reprs = collections.OrderedDict()
-        newish_local_reprs = collections.OrderedDict()
-
-        for key, value in local_reprs.items():
-            if key not in old_local_reprs:
-                newish_local_reprs[key] = value
-            elif old_local_reprs[key] != value:
-                modified_local_reprs[key] = value
-
         newish_string = ('Starting var:.. ' if event == 'call' else
                                                             'New var:....... ')
-        for name, value_repr in newish_local_reprs.items():
-            self.write('{indent}{newish_string}{name} = {value_repr}'.format(
+
+        for name, value_repr in local_reprs.items():
+            if name not in old_local_reprs:
+                self.write('{indent}{newish_string}{name} = {value_repr}'.format(
+                                                                       **locals()))
+
+        for name, value_repr in local_reprs.items():
+            if name in old_local_reprs and old_local_reprs[name] != value_repr:
+                self.write('{indent}Modified var:.. {name} = {value_repr}'.format(
                                                                    **locals()))
-        for name, value_repr in modified_local_reprs.items():
-            self.write('{indent}Modified var:.. {name} = {value_repr}'.format(
-                                                                   **locals()))
+
         #                                                                     #
         ### Finished newish and modified variables. ###########################
 
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -19,13 +19,14 @@ ipython_filename_pattern = re.compile('^<ipython-input-([0-9]+)-.*>$')
 
 
 def get_local_reprs(frame, watch=()):
+    var_names = frame.f_code.co_varnames + frame.f_code.co_cellvars
     result = collections.OrderedDict(
         (key, utils.get_shortish_repr(frame.f_locals[key]))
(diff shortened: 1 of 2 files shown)


pysnooper/tracer.py
```diff
@@ -19,10 +19,16 @@
...
 def get_local_reprs(frame, watch=()):
-    result = {key: utils.get_shortish_repr(value) for key, value
-                                                     in frame.f_locals.items()}
+    result = collections.OrderedDict(
+        (key, utils.get_shortish_repr(frame.f_locals[key]))
+        for key in frame.f_code.co_varnames if key in frame.f_locals
+    )
+
+    result.update(sorted((key, utils.get_shortish_repr(frame.f_locals[key]))
+                  for key in set(frame.f_locals) - set(frame.f_code.co_varnames)))
```
alexmojaki: How about: 

```python
keys = [
    key for key in frame.f_code.co_varnames if key in frame.f_locals
] + sorted(set(frame.f_locals) - set(frame.f_code.co_varnames))

result = collections.OrderedDict(
    (key, utils.get_shortish_repr(frame.f_locals[key]))
    for key in keys
)
```

pysnooper/tracer.py
```diff
@@ -182,7 +188,7 @@ def __init__(
              v if isinstance(v, BaseVariable) else Exploding(v)
              for v in utils.ensure_tuple(watch_explode)
         ]
-        self.frame_to_local_reprs = collections.defaultdict(lambda: {})
+        self.frame_to_local_reprs = collections.OrderedDict()
```
alexmojaki:
```suggestion
        self.frame_to_local_reprs = {}
```

I agree there's not much use to a defaultdict here (we'll see what @cool-RR thinks) but there's no reason for this to be ordered.

pysnooper/tracer.py
```diff
@@ -287,10 +293,10 @@ def trace(self, frame, event, arg):
 
         newish_string = ('Starting var:.. ' if event == 'call' else
                                                             'New var:....... ')
-        for name, value_repr in sorted(newish_local_reprs.items()):
+        for name, value_repr in newish_local_reprs.items():
```
alexmojaki: This can be taken a step further. Since we're not sorting, we don't need intermediate dicts like `newish_local_reprs`. We can just write immediately in the loop above. Kills several lines of code.

alexmojaki:
Overall I think this is a great idea and good work! It might be nice to add a test matching your demo, but it's not critical.

pysnooper/tracer.py
cool-RR: Agreed there's no need for `defaultdict`. I'll let you guys figure out whether an `OrderedDict` is necessary or not, I don't mind.

pysnooper/tracer.py
cool-RR: Agreed.

pysnooper/tracer.py
cool-RR: Both of your attempts are ugly in my opinion. Let me try...

alexmojaki: It looks like `sorted(set(frame.f_locals) - set(frame.f_code.co_varnames))` can be replaced by `frame.f_code.co_freevars`:

```python
def foo():
    y = 2
    x = 1

    def bar():
        print(y, x)

    print(bar.__code__.co_freevars)

foo()
```

But I don't know for sure.

pysnooper/tracer.py
alexbers: Done, https://github.com/cool-RR/PySnooper/pull/113/commits/d99921f7fd593714643b975d1c8ea254325e0ffc

pysnooper/tracer.py
```diff
@@ -272,27 +278,23 @@ def trace(self, frame, event, arg):
...
-            self.write('{indent}Modified var:.. {name} = {value_repr}'.format(
+
+        for name, value_repr in local_reprs.items():
+            if name not in old_local_reprs:
+                self.write('{indent}{newish_string}{name} = {value_repr}'.format(
+                                                                       **locals()))
+
+        for name, value_repr in local_reprs.items():
+            if name in old_local_reprs and old_local_reprs[name] != value_repr:
+                self.write('{indent}Modified var:.. {name} = {value_repr}'.format(
```
alexmojaki: This is actually not what I had in mind. I guess you're trying to make sure that new variables are all written before modified ones, which I hadn't thought about. I'm not actually sure if that's a constraint worth sticking to. Maybe variable declaration order should always win?

cool-RR:
Good work. Please cherrypick my commit 9a2deda7fd63a579302989dfdb593836f8a1a1eb. The code for getting the variables in the right order is still ugly, but now the ugliness is confined to the sorting. This way it's easier for people to understand what the actual items are without delving into the sorting code.

pysnooper/tracer.py
cool-RR: I don't mind either way. Alex, feel free to make a decision here.

pysnooper/tracer.py
cool-RR: I wrote my own version in `9a2deda7fd63a579302989dfdb593836f8a1a1eb`, please cherrypick that. The code for getting the variables in the right order is still ugly, but now the ugliness is confined to the sorting. This way it's easier for people to understand what the actual items are without delving into the sorting code.

alexbers: Done, https://github.com/cool-RR/PySnooper/pull/113/commits/a2d848245ec2f7b4328e98d8e53ee25737c235a6.

Also I specified the the order explicitly:
    
    vars_order = frame.f_code.co_varnames + frame.f_code.co_cellvars + frame.f_code.co_freevars
 
Also, if we don't know how to sort the string, we can sort it alphabetically instead of making the tuple from dict keys. This makes code a bit simpler, faster, and the result is more predictable between python implementations. Actually, on cpython there shouldn't be any variables other than in co_varnames, co_cellvars or co_freevars

pysnooper/tracer.py
```diff
@@ -19,10 +19,21 @@
 
 
 def get_local_reprs(frame, watch=()):
-    result = {key: utils.get_shortish_repr(value) for key, value
-                                                     in frame.f_locals.items()}
+    result_items = [(key, utils.get_shortish_repr(value)) for key, value
+                                                     in frame.f_locals.items()]
+
+    vars_order = frame.f_code.co_varnames + frame.f_code.co_cellvars + frame.f_code.co_freevars
```
alexmojaki:
```suggestion
    vars_order = frame.f_code.co_varnames + frame.f_code.co_freevars
```

`co_cellvars` names variables in functions inside this one, we don't care about it.

pysnooper/tracer.py
alexmojaki: OK, remove the ordereddict. we don't care about the order of frames.

pysnooper/tracer.py
alexmojaki: OK, let's make variable declaration order king. Flatten the code into a single loop.

alexbers: Yes, the variable declaration order still wins, because the first message is always about the declaration and the further ones - about modifying.
The only exception I know if when we use watch. For example if we have a list of [7,8,9] and append 10 to it we can get:

    Modified var:.. lst = [7, 8, 9, 10] 
    New var:....... (lst + [])[3] = 10
    Modified var:.. lst + [] = [7, 8, 9, 10]

I added a test on ordering and another test on cell variables in nested functions.

alexmojaki: Are you saying that sample output is what you currently get, or what you would get if you changed the code? Is it good or bad?

pysnooper/tracer.py
alexbers: Here is an example why we should:

```python
import pysnooper

@pysnooper.snoop(depth=2)
def f1(a):
    def f2(a):
        x = 0
        x += 1
        def f3(a):
            y = x
            return 42
        return f3(a)
    return f2(a)

f1(1)
```

The x variable is in co_cellvars.
If we try sort with this simpler function, which ignores the unknown vars:
```python
result_items.sort(key=lambda key_value: vars_order.index(key_value[0]))
```
... we can get that is works if we use co_cellvars and doesn't work if we don't.

I think this simple function should always work on cpython, but I don't sure about other implementations.

alexmojaki: I checked out your repo and got the exact same results with your sample script with and without my change.

> If we try sort with this simpler function

but we're not, are we? I agree it's best to avoid relying too much on the workings of code objects.

cool-RR: Your `vars_order` list is a great idea. How about you add f_locals to the end of it? Even if some variables repeat themselves, that wouldn't hurt. That way we'll be 100% sure no variable makes it out of the order, no matter what magic the user used when defining his function.

cool-RR: And of course, that way we could remove this ugly check: ` if key_value[0] in vars_order else float("inf"))`

alexbers: Could you please repeat the test, because for me it is reproducible on Python 3.5, 3.6 3.7 and on PyPy/PyPy3.

The simpler function was only for example where co_cellvars considering is needed.

pysnooper/tracer.py
alexbers: Currently the new vars are always before the modified ones. This is corresponds to the old behavior (before this pull request).

For me the both variants seems good.

pysnooper/tracer.py
alexbers: > Your `vars_order` list is a great idea. How about you add f_locals to the end of it? Even if some variables repeat themselves, that wouldn't hurt. That way we'll be 100% sure no variable makes it out of the order, no matter what magic the user used when defining his function.

Done, https://github.com/cool-RR/PySnooper/pull/113/commits/00591f658859eb3c6c8fa33e0973bbb0368d0a06. The code in get_local_reprs became much nicer to read.

pysnooper/tracer.py
alexmojaki: Right, I'm saying let's drop that old behaviour. New and modified can be in any order, I don't think it improves anyone's experience, @cool-RR seems to be saying it's fine. That's what I mean by variable declaration order is king, it's the only consideration, and I think that's good for both the code and the user.

alexbers: Ok, done, https://github.com/cool-RR/PySnooper/pull/113/commits/21d59bc6120f6072ac3656e16e55948424e3a1b3

cool-RR:
Good job. @alexmojaki Is there anything else you need changed in this PR before it's merged? If so can you give me a list?

pysnooper/tracer.py
```diff
@@ -19,10 +19,14 @@
 
 
 def get_local_reprs(frame, watch=()):
-    result = {key: utils.get_shortish_repr(value) for key, value
-                                                     in frame.f_locals.items()}
+    vars_order = frame.f_code.co_varnames + frame.f_code.co_cellvars + frame.f_code.co_freevars + tuple(frame.f_locals.keys())
```
alexmojaki: OK, I thought that co_cellvars must always be a subset of co_varnames, but I see that it's not. Which is really weird.

```suggestion
    code = frame.f_code
    vars_order = code.co_varnames + code.co_cellvars + code.co_freevars + tuple(frame.f_locals.keys())
```

alexmojaki:
I think once tests pass this is good to go.

cool-RR:
Woohoo! Merged and made a new release with this feature.

Nice work @alexbers , we grilled you but you made it through :)

PR #114, opened 2019-05-11
Log exceptions nicely

alexmojaki:
Script:

```python
import pysnooper


@pysnooper.snoop()
def foo():
    1 / 0


@pysnooper.snoop()
def bar():
    foo()


bar()
```

Before:

```
08:58:06.538289 call        10 def bar():
08:58:06.538506 line        11     foo()
    08:58:06.538567 call         5 def foo():
    08:58:06.538613 line         6     1 / 0
    08:58:06.538650 exception    6     1 / 0
    08:58:06.538685 return       6     1 / 0
    Return value:.. None
08:58:06.538751 exception   11     foo()
08:58:06.538786 return      11     foo()
Return value:.. None
Traceback (most recent call last):
```

After:

```
09:04:08.579983 call        10 def bar():
09:04:08.580215 line        11     foo()
    09:04:08.580272 call         5 def foo():
    09:04:08.580315 line         6     1 / 0
    09:04:08.580352 exception    6     1 / 0
    ZeroDivisionError: division by zero
    Call ended by exception
09:04:08.580491 exception   11     foo()
ZeroDivisionError: division by zero
Call ended by exception
Traceback (most recent call last):
```

### DIFF
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -339,10 +339,13 @@ class Tracer:
         # If a call ends due to an exception, we still get a 'return' event
         # with arg = None. This seems to be the only way to tell the difference
         # https://stackoverflow.com/a/12800909/2482744
+        code_byte = frame.f_code.co_code[frame.f_lasti]
+        if not isinstance(code_byte, int):
+            code_byte = ord(code_byte)
         ended_by_exception = (
                 event == 'return'
                 and arg is None
-                and (opcode.opname[frame.f_code.co_code[frame.f_lasti]]
+                and (opcode.opname[code_byte]
                      not in ('RETURN_VALUE', 'YIELD_VALUE'))
         )
 
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -3,12 +3,14 @@
 
 import functools
 import inspect
+import opcode
 import sys
 import re
 import collections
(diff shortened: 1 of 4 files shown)


cool-RR:
This is nice. However, the trace function is already a big spaghetti already, so I'd appreciate if any additions would be easy to understand at a glance. If you're adding code that's relevant only to `return` calls, I think it should all be under some kind of `if event == 'return'` so it'll be easy for a casual reader to say "Oh, that's relevant only for return calls, I can skip that right now."

alexmojaki:
I can't do that, because it determines whether the main event line is written at all.

cool-RR:
I understand, that's okay.

cool-RR:
Ready as far as you're concerned?

alexmojaki:
Once the new tests pass on Travis, yes.

This style of tests is much nicer to work with. Much easier and faster to read, write, and maintain. No regexes. Much less thinking and typing required. Much more concrete, no abstractions. I think it would be a huge improvement to migrate all tests related to tracing to this style. Meanwhile tests relating to output methods should focus only on testing the output, similar to what I've done in #103. We can get rid of `assert_output` and the entries framework entirely.

cool-RR:
Interesting, please move the discussion about tests to a separate issue.

Issue #115, opened 2019-05-11
Simplify tests using `assert_sample_output`, no more entries and regexes


alexmojaki:
Tests using `assert_sample_output` are much easier and faster to read, write, and maintain. No regexes. Much less thinking and typing required. Much more concrete, no abstractions. I think it would be a huge improvement to migrate all tests related to tracing to this style. Meanwhile tests relating to output methods should focus only on testing the output, similar to what I've done in #103. We can get rid of `assert_output` and the entries framework entirely.

cool-RR:
Interesting. Can you give a list of disadvantages that these tests would
have over the entries framework?

alexmojaki:
In some scenarios a change to the implementation might result in a large batch of tests needing updating that may not be needed in the entries framework. This is pretty simple mindless copy-pasting, so I don't think it's a problem. I could probably write a script to automate it. And I think it would be good to have a clear diff showing how a change affects the output in various situations.

Occasionally, not being able to specify things like value_regex might make things a little harder, but not much. Some regexes in the normalise function instead, maybe a bit of patching, and it should be fine.

cool-RR:
Okay, thanks. It might take me a while to get to this.

alexmojaki:
I'm willing to do this.

cool-RR:
I understand, that's good. I'll need a while to think about it anyway, so I
suggest you wait until I update here.

PR #117, opened 2019-05-11
Only test writing in tests about writing

alexmojaki:
Taken from #103, one step towards #115. There's no reason to test actual tracing just to make sure that the various write functions work.

### DIFF
--- a/tests/test_pysnooper.py
+++ b/tests/test_pysnooper.py
@@ -4,7 +4,6 @@
 import io
 import textwrap
 import threading
-import types
 import sys
 
 from pysnooper.utils import truncate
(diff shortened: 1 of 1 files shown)


cool-RR:
I don't like this. I prefer tests to be mixing things up, being end-to-end.

alexmojaki:
The other tests do that, with stderr. There is no incorrect implementation that will leave these tests passing. One of the two sets must fail.

It's normal to have some tests that aren't end to end, those are unit tests. We have others already.

cool-RR:
Sorry, I like the current tests. It's valuable to see that snooping still works when you're using different arguments.

PR #118, opened 2019-05-11
Allow using with block without parentheses

alexmojaki:
This lets you write:

    with pysnooper.snoop:

It's a pretty silly idea, and it's fine if you reject it, but it was kinda fun to write.


cool-RR:
Cool idea and implementation, but yeah, not worth the complexity.

Issue #104, opened 2019-05-09
How to display this string completely?


kagxin:
![image](https://user-images.githubusercontent.com/16145883/57423260-d72ef700-7245-11e9-9b10-b8b6b5146afc.png)
How to display this string completely?

cool-RR:
@alexmojaki Why is this string showing so short? When I wrote the code, I made it so it'll truncate at around 100 characters, and when you refactored, you set `self.maxother = 100`. I guess `reprlib.Repr` works differently? In any case I definitely don't want strings to be truncated so much. Can this be fixed?

alexmojaki:
There's a separate setting `maxstring`.

As for how to display a specific variable completely, I would recommend `print`.

cool-RR:
In your implementation, is there a way to accomplish the following? "Whenever `repr(x)` would be under 100 characters, display it in full without any truncation"

alexmojaki:
No, there's no way to tell if e.g. a list is going to have a long repr without just computing the full thing.

cool-RR:
I understand. I removed `reprlib` and switched back to `repr`. I don't want to sacrifice simple, common usecases for more esoteric ones.

@kagxin Please upgrade PySnooper. Now your string will display, assuming its representation is below 100 characters. If it's above 100 characters, you can monkeypatch PySnooper like this `pysnooper.utils.MAX_VARIABLE_LENGTH = 1000`

alexmojaki:
Well you could still compute the full `repr` and switch to `reprlib` if it's too long. There are advantages besides performance.

> I don't want to sacrifice simple, common usecases for more esoteric ones.

Keep in mind that this issue was not such a use case, just a reminder that we should have set `maxstring` as well.

I don't know which use cases you have in mind, it's only under pretty specific circumstances that your implementation will display information the user wants that `reprlib` won't and that wouldn't be better obtained by other means.

kagxin:
thanks

PR #119, opened 2019-05-11
Add log function

alexmojaki:
I know this library is against print style debugging, but sometimes it has its uses. This helps when the usual pysnooper methods are not ideal. For example:

1. To see a single full value, as in #104. This is more convenient than changing MAX_VARIABLE_LENGTH and doesn't globally make the output more verbose.
2. To see just one value without all the other stuff from `snoop`. Even using `with` on a single line can print out a bunch of irrelevant variables. This can be used in places where there is no decorator or block.
3. To see a value that doesn't have a local variable, without having to rearrange the code and invent a temporary variable. Like the libraries q and icecream, this lets the arguments pass through for minimal code disruption. It's also good for people who are tired and used to print debugging so they don't think of adding a local variable.
4. To see a value just once, if evaluating its repr every line is terrible for performance or seeing every modification to its value would be disruptive.

The main advantage over `print` is that this is combined with the other output nicely, no matter the output method.

### DIFF
--- a/pysnooper/__init__.py
+++ b/pysnooper/__init__.py
@@ -17,7 +17,7 @@ changed in the decorated function.
 For more information, see https://github.com/cool-RR/PySnooper
 '''
 
-from .tracer import Tracer as snoop
+from .tracer import Tracer as snoop, log
 from .variables import Attrs, Exploding, Indices, Keys
 import collections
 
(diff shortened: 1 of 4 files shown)


cool-RR:
This is nice functionality, but I'm very hesitant to increase the scope of PySnooper. It's supposed to be a minimalistic little tool, and adding stuff might hurt that. Sorry. People who want to print their expressions can just print them, and people who want a powerful debugger can use an actual debugger.

PR #122, opened 2019-05-13
Get rid of the six module

alexbers:
This pull requests shrinks the code base three times, from 1 500 lines to 500 by removing the six third party module. The 1 000 removed lines of code are the mostly code for old python version compatibility, which can influence the running program. For example, the six module patches the import system adding its importer to sys.meta_path.

The PySnooper still works on Python 2.7.


cool-RR:
Great idea Alex! I'm happy to get rid of all that unnecessary code.

PR #130, opened 2019-06-17
Support activate and desactivate by environment and configure snoop globally

pikez:


pysnooper/tracer.py
```diff
@@ -400,3 +408,9 @@ def trace(self, frame, event, arg):
                        format(**locals()))
 
         return self.trace
+
+    @classmethod
+    def setup(cls, **kwargs):
+        if not (set(pycompat.getfullargspec(cls.__init__).args) > set(kwargs.keys())):
+            raise Exception('The parameters passed to setup contain non-snoop parameters')
+        return functools.partial(cls, **kwargs)
```
pikez: the setup function can configure globally. I don't know if you like this.

pikez: the setup function can configure globally. I don't know if you like this.

cool-RR:
Good work. I indeed don't like the setup function, please remove it. Instead of an attribute `disable` it should be a top-level constant `DISABLED`.

pikez:
ok. I removed it. It's more concise now.

tests/test_pysnooper.py
```diff
@@ -1172,3 +1173,21 @@ def sum_to_x(x):
...
+def test_activate_deactivate_snoop():
+    string_io = io.StringIO()
+
+    def my_function(foo):
+        x = 7
+        y = 8
+        return x + y
+
+    os.environ['PYSNOOPER_DISABLED'] = '1'
+    with pysnooper.snoop():
```
alexmojaki:
```suggestion
    with pysnooper.snoop(output):
```

Right now you're not actually testing anything.

tests/test_pysnooper.py
```diff
@@ -1172,3 +1173,21 @@ def sum_to_x(x):
...
+
+def test_activate_deactivate_snoop():
+    string_io = io.StringIO()
+
+    def my_function(foo):
+        x = 7
+        y = 8
+        return x + y
+
+    os.environ['PYSNOOPER_DISABLED'] = '1'
```
alexmojaki: This won't work because DISABLED is a constant. You need something like:

```suggestion
    pysnooper.tracer.DISABLED = '1'
```

tests/test_pysnooper.py
pikez: yes! I'm wrong. it's should pass the `string_io` object. I'll change it.

cool-RR:
@pikez Good work! I merged this manually now and made a new release with this feature.

PR #148, opened 2019-08-09
Show source path, especially when multiple files

cool-RR:


cool-RR:
@alexmojaki What do you think about this? I needed this functionality at work the other day.

alexmojaki:
Be aware that this may show the source path at times other than the start of a call/block if multiple threads interleave. Maybe that's a good thing.

`last_source_path` should probably be a global variable, since each `@snoop()` makes a new `Tracer`.

alexmojaki:
Let me clarify the second point. Suppose you have functions `a` and `b` in two files, both snooped, and you call `a(); b(); a()`. You probably want the source file to show every time, particularly on the second call to `a` when it changed from `b.py` to `a.py`. If I'm reading correctly, that won't happen, because the Tracer for `a` will still think the last tracing happened in `a.py`.

cool-RR:
Correct. We have a general problem with the architecture, regarding putting state on the tracer. Since it's a class, it's implicit that each Tracer should have its state independent. Especially considering you can give different settings to different tracers. But the problem you raised is indeed a problem.

Of course, if I changed it to a global, then your problem would be fixed, but what if someone makes two tracers, each to a separate file? Then in your approach, there would be redundant lines about the source code in each dump.

So I'm gonna leave things as is, because all of these problems are mostly esoteric, and I'll keep assuming that people use PySnooper is mostly simple ways, and that giving too much energy to complex scenarios is not worth it.

alexmojaki:
Right, this is one reason why I keep my configuration such as the output file separate from the decorator calls. Then state like this is tied to one such configuration, so it's almost global but avoids the problem you mentioned. Plus I think it's weird to make the user copy something like `@snoop('/tmp/snoop.log', prefix='zzz', custom_repr=(...))` to each function that they want to trace with the same options.

> I'll keep assuming that people use PySnooper is mostly simple ways, and that giving too much energy to complex scenarios is not worth it.

But you're the one worrying about a complex scenario, namely outputting to two files. My scenario is just two functions in two different files being traced, which is actually also a necessary part of your scenario, otherwise there wouldn't be a problem about redundant lines.

cool-RR:
Yes, I guess I'm just wary of moving state out of the class, even if just
for appearance.

alexmojaki:
By the way, have you ever used snoop in practice? I think it would be helpful to both of us to use the other library now and then, and this would have been a great opportunity when you "needed this functionality at work the other day" since snoop offers it in two forms.

cool-RR:
The usage that led to this ticket, a couple of weeks ago, is actually my first usage of PySnooper :D I use a real debugger when I work. My debugger had a bug so I had to debug it. 

Relevant :) https://www.youtube.com/watch?v=APl0edZ8gI8

I don't think I'll use PySnooper often so I'm not likely to make time to use snoop.

alexmojaki:
Haha wow, that's a bit crazy that you haven't used it before. I'll admit that I haven't either, but in my defense I have a version which I prefer and which I do use now and then. I also use my IDE debugger much more than snoop or birdseye, but there are definitely situations where they're useful.

Ever heard of dogfooding? You should probably make an effort to use PySnooper once in a while even when you don't feel that it's the best tool for the job, just to get a feel for it.

cool-RR:
Yes, I agree.

PR #149, opened 2019-08-23
Add truncate option support

deagon:


pysnooper/tracer.py
```diff
@@ -185,6 +188,14 @@ class Tracer:
 
         @pysnooper.snoop(custom_repr=((type1, custom_repr_func1), (condition2, custom_repr_func2), ...))
 
+    Customize whether truncate the variable value or not::
+
+        @pysnooper.snoop(truncate=False)
```
alexmojaki: This doesn't match the `need_truncate` parameter.

pysnooper/utils.py
```diff
@@ -58,14 +55,16 @@ def get_repr_function(item, custom_repr):
...
-def get_shortish_repr(item, custom_repr=()):
+def get_shortish_repr(item, custom_repr=(),
+                      need_truncate=True, max_length=100):
     repr_function = get_repr_function(item, custom_repr)
     try:
         r = repr_function(item)
     except Exception:
         r = 'REPR FAILED'
     r = r.replace('\r', '').replace('\n', '')
-    r = truncate(r, MAX_VARIABLE_LENGTH)
```
alexmojaki: I'm noticing a potential problem that's related to this PR but not caused by it. If someone provides a custom repr, it can still get truncated. I think a custom repr should override that and be shown in full.

deagon: Yes. I've noticed that too.
However, user could pass "need_truncate=False" when he wants to use custom repr without truncated.
This option is useful for user to decide what he wants.

pysnooper/tracer.py
deagon: Good catch.

cool-RR:
@deagon Congrats on your first contribution! Thank you for working on this. 

I've given this PR some thought, and I'm conflicted. On one hand, it's a nice feature. On the other hand, I think that it's adding way too much complexity to the code. I'd like to keep PySnooper as a cute toy rather than a debugger.

Here is my proposed solution: Instead of 3 arguments, `need_truncate`, `max_exception_length` and `max_variable_length` there will be only one argument, which is `max_variable_length`. The same length number will be used for truncating exceptions too.

If given `max_variable_length=None`, then no truncation will be done.

Regarding the custom repr overriding the max variable length: I understand the reasoning but I disagree. The max variable length is there to protect from a massive text dump in some cases that could result in a big mess. People who are fancy enough to use a custom repr, which is a tiny use case, are also fancy enough to set the `max_variable_length` argument.

Assuming you can get that implemented and cut down on a lot of the complexity of this PR, then after I review it, I'll ask you to also mention it in the docstring and the readme, and add tests.

deagon:
@cool-RR Agree with you.
I'll fix it later. And I'm glad to contribute on this wonderful tool.

cool-RR:
Let me know when it's ready for review.

cool-RR:
@deagon Don't give up, I believe in you :)

alexmojaki:
Just for comparison, [snoop](https://github.com/alexmojaki/snoop):

- never truncates the exception
- lets you configure the verbosity of each type separately via `cheap_repr`
- doesn't have a way to increase the verbosity of all types

deagon:
@cool-RR Sorry for my late reply. 
I was lost in another work. I'll fix it today or tomorrow.

cool-RR:
Crocodile

deagon:
@cool-RR 
Sorry again for my late. You can review it now.

cool-RR:
@deagon Good job! I merged your commit manually, edited some stuff and made a release. Congrats on your first contribution!

cool-RR:
@deagon Followup on this: I just used this feature at work today, and it was very helpful :)

deagon:
@cool-RR Hah! I got it.

Issue #150, opened 2019-09-07
Allow snooping on all methods of a class


jonathangjertsen:
I didn't find a way to easily snoop on all the methods of a class without adding the decorator to every single method. Below is an outline of a way to do it with a metaclass. There was an issue overwriting the magic/dunder methods (`AttributeError: type object 'SnoopedMetaClass' has no attribute '__code__'`), but for everything else it works fine.

```
import pysnooper

class SnoopMetaClass(type):
    """Converts every method to a snooped method"""
    def __new__(cls, name, bases, attrs, **kwargs):
        # Create the instance like normal
        instance = super().__new__(cls, name, bases, attrs)

        # If the class has a "snoop_kwargs" class attribute, get it here
        # so we can forward it to pysnooper.snoop. Set it to an empty dict by default.
        kwargs = attrs.get("snoop_kwargs", {})

        # Iterate over every attribute and find the ones that are methods
        for attr in dir(instance):
            # Certain magic methods do not like to be overridden by the snooped version
            if attr.startswith("__"):
                continue

            # Override the value with a snooped one if it is callable
            value = getattr(instance, attr)
            if callable(value):
                snooped = pysnooper.snoop(**kwargs)(value)
                setattr(instance, attr, snooped)
        return instance




class MyClass(metaclass=SnoopMetaClass):
    snoop_kwargs = {"output": "log.txt"}

   def __init__(self):
       """Will not be snooped on since it is a magic method"""

    def my_method(self):
        """Normal method, will be snooped on"""

    def my_other_method(self):
        """Normal method, will be snooped on"""
```

cool-RR:
This is a cute idea. My priority is to keep the interface as simple as possible, so if you'd like me to merge it, I'd like you to modify it so it's a decorator on the class rather than a metaclass. Submit a PR like that, with tests and documentation, and I'll likely merge it.

alexmojaki:
I considered doing this for snoop, I just wanted to hear other people's thoughts on it first: https://github.com/alexmojaki/snoop/issues/2

The main problem with the kind of implementation above (metaclass or not) is that decorated methods in the class won't be traced correctly. If you have a traditional decorator:

```python
def my_decorator(f):
    def wrapper(*args, **kwargs):
        ...
    return wrapper
```

And a method:

```python
@my_decorator
def my_method
```

[...]

PR #151, opened 2019-09-08
Allow snooping on all methods of a class (issue #150)

jonathangjertsen:
This implements https://github.com/cool-RR/PySnooper/issues/150 by allowing the same decorator to be applied to a decorator or to a class.

Based on input from @alexmojaki I have added tests to verify that `snoop` at least does not actively break properties or other methods with decorators. The output is probably suboptimal/confusing in those cases, but I think that should be OK given the warning in README.md.


tests/test_pysnooper.py
```diff
@@ -1237,3 +1237,214 @@ def my_function(foo):
...
+    assert_output(
+        output,
+        (
+            SourcePathEntry(),
+            VariableEntry('self', value_regex="u?.*<locals>.MyClass object at"),
+            CallEntry('def __init__(self):'),
+            LineEntry('self.x = 7'),
+            ReturnEntry('self.x = 7'),
+            ReturnValueEntry('None'),
+            VariableEntry('args', value_regex="u?.*"),
```
jonathangjertsen: The corresponding line is

```
Starting var:.. args = (<tests.test_pysnooper.test_class_with_decorated...s>.MyClass object at 0x0000028E03C69048>, 'baba')
```
Not sure what would constitute a sensible regex here...

pysnooper/tracer.py
```diff
@@ -223,9 +223,23 @@ def __init__(
...
         if DISABLED:
             return function
+
+        if inspect.isclass(function_or_class):
+            return self._wrap_class(function_or_class)
+        else:
+            return self._wrap_function(function_or_class)
+
+    def _wrap_class(self, cls):
+        for attr_name in dir(cls):
```
alexmojaki: This will snoop methods from base classes, which again can lead to logs from snoop even when the decorated class isn't being used. I suggest using `__dict__` to only touch methods defined directly in the current class.

tests/test_pysnooper.py
alexmojaki: `r"\(<.+>, 'baba'\)"`

alexmojaki:
> I think that should be OK given the warning in README.md.

Try not to rely on documentation. At least some users are going to try snooping a class without ever seeing it mentioned in the README. Others will see the warning but it may only be relevant to them much later and they will have forgotten.

pysnooper/tracer.py
cool-RR: Agreed.

cool-RR:
Looking good. I like the extensive tests. I'm gonna edit your code a little bit after I accept it, so you just get this PR to the point where the tests pass, and I'll do the rest.

jonathangjertsen:
> Try not to rely on documentation. At least some users are going to try snooping a class without ever seeing it mentioned in the README. Others will see the warning but it may only be relevant to them much later and they will have forgotten.

Could this be prevented from having a separate name for the class decorator? `@pysnooper.snoop_on_all_methods_in_this_class_except_methods_which_have_been_decorated_in_which_case_the_behaviour_depends_on_the_type_of_the_return_value_of_the_decorator` is not so catchy, but maybe something to the same effect.

alexmojaki:
You can't catch all cases but you can add code that accounts for common decorators (property, classmethod, staticmethod, lru_cache) and decorates their underlying functions.

jonathangjertsen:
> You can't catch all cases but you can add code that accounts for common decorators (property, classmethod, staticmethod, lru_cache) and decorates their underlying functions.

Looking at this in a separate branch, perhaps a goal can be to account for all of the builtins or even all of the ones in the standard library. It's gonna end up as a lot of code though, at least 3 lines per decorator.

cool-RR:
Guys, remember that this is an esoteric feature as it is. We definitely
don't need to support property in this PR, I doubt that we should even
support it in a separate PR.

alexmojaki:
I don't think it's such an esoteric feature, I think many people naturally consider this possibility. The same thing happened with birdseye.

Supporting property and a few others is not hard and adds a lot of value. Supporting everything in the standard library is probably pushing it. In any case it may not be a good idea to touch decorators that return normal functions. @jonathangjertsen do you have any other examples of decorators to handle?

By the way, we could also easily decorate the `__call__` methods of user defined classes whenever an instance is snooped. That might even be useful outside of this PR.

jonathangjertsen:
@cool-RR I don't plan on doing anything more with this PR now that the tests are passing.

@alexmojaki There are quite a few more decorators mentioned here https://github.com/lord63/awesome-python-decorator

I'm experimenting with support for some decorators in https://github.com/cool-RR/PySnooper/compare/master...jonathangjertsen:manual-decorator-unwrapping, but it's mostly for learning.

alexmojaki:
OK I looked through those, I don't think it would be sensible to specially support any of them.

However you should make sure you skip anything that passes `inspect.iscoroutinefunction` or the whole thing will fail.

jonathangjertsen:
Good catch, I've pushed a fix, I didn't find a way to write a test that didn't cause a syntax error on older Python versions though.

cool-RR:
Great job Jonathan. I merged your commits manually, fixed your bug :) , edited the docs some and made a release.

PR #169, opened 2019-11-19
Feature/normalize output

itraviv:
normalizing output - removing machine specific data:
- absolute paths are replaced with the `basename` of the path.
- time stamps are removed.
- default (CPython) reprs of the form <.... at 0x[0-9a-fA-F] > will not contain the 'at 0x...' ending.

tests:
- normalization does not affect the original test results so every test now runs with `normalize=True`
and `normalize=False` to preserve compatability.
- `assert_output` can now verify the output of a normalized trace.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ build/
 
 .coverage
 htmlcov
+/venv/
+.idea
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -15,9 +15,9 @@ import traceback
 
 from .variables import CommonVariable, Exploding, BaseVariable
 from . import utils, pycompat
+
 if pycompat.PY2:
     from io import open
-
 
 ipython_filename_pattern = re.compile('^<ipython-input-([0-9]+)-.*>$')
 
@@ -73,7 +73,7 @@ def get_path_and_source_from_frame(frame):
                 import IPython
                 ipython_shell = IPython.get_ipython()
                 ((_, _, source_chunk),) = ipython_shell.history_manager. \
-                                  get_range(0, entry_number, entry_number + 1)
+                    get_range(0, entry_number, entry_number + 1)
                 source = source_chunk.splitlines()
             except Exception:
                 pass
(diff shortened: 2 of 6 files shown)


cool-RR:
Good job. Before I do a thorough code review, please take care of these issues:

1. Your code shouldn't be changing whitespace of any lines that aren't related to your code. Go over the "Files changed" in this PR and you'll see that there are a bunch of changes like that. 
2. I see that you check whether the default Python repr was used by using a regular expression. That's an okay solution, but if you can instead do a check that doesn't rely on regexp, that'd be better.
3. It would be nicer if your commits were more finalized (probably less commits, clearer messages), but I won't hold up your PR for that, worst-case scenario I'll organize them later.

I see that you closed your PR, maybe you're opening another one or something, I'll stay tuned.

itraviv:
i closed it because it failed the CI. i will fix and organize the commits as you requested.
stay tuned for an improved version.

cool-RR:
Cool. You could keep it open if you want though and keep updating it until
the CI passes, it's all good. You can also run tests locally with pytest, I
think that just running `pytest` will do the trick.

itraviv:
yea, pytest is not the issue here (i have also added tests), it's the python2.7 compatibility that fails.
as for the regex - i couldn't find a better way since the number of figures appearing in the memory address (after the 0x..) is very platform and context specific (memory address range). if you have any other suggestion - please share with me.

cool-RR:
Hmm, I think it's possible to figure out whether the default repr is used, here's some inspiration under Python 3, you could probably do something similar in Python 2:

```
In [15]: class A: 
    ...:     pass 
    ...: 
 
In [16]: class B(A): 
    ...:     def __repr__(self): return 'foo' 
    ...: 
 
In [17]: a.__repr__ 
Out[17]: <method-wrapper '__repr__' of A object at 0x0000000004D0D748> 
 
In [18]: a.__repr__.__objclass__ 
Out[18]: object 
 
In [19]: b.__repr__ 
Out[19]: <bound method B.__repr__ of foo> 
 
In [20]: b.__repr__.__func__ 
Out[20]: <function __main__.B.__repr__(self)> 
 
In [21]: b.__repr__.__func__.__qualname__ 
Out[21]: 'B.__repr__' 
```

Another thought though: Maybe there are non-default reprs which also use the memory address. I actually think your regex method might be superior, so let's go with that. I'm keeping my above code here for reference in case we'll want it later.

itraviv:
opened new PR with all mentioned changes: https://github.com/cool-RR/PySnooper/pull/170

PR #170, opened 2019-11-19
Feature/normalize output

itraviv:
normalizing output - removing machine specific data:

    absolute paths are replaced with the basename of the path.
    time stamps are removed.
    default (CPython) reprs of the form <.... at 0x[0-9a-fA-F] > will not contain the 'at 0x...' ending.

tests:

    normalization does not affect the original test results so every test now runs with normalize=True
    and normalize=False to preserve compatability.
    assert_output can now verify the output of a normalized trace.

readme:
- updated accordingly

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,12 @@ Start all snoop lines with a prefix, to grep for them easily:
 @pysnooper.snoop(prefix='ZZZ ')
 ```
 
+Remove all machine-related data (paths, timestamps, memory addresses) to compare with other traces easily:
+
+```python
+@pysnooper.snoop(normalize=True)
+```
+
 On multi-threaded apps identify which thread are snooped in output:
 
 ```python
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -22,19 +22,19 @@ if pycompat.PY2:
 ipython_filename_pattern = re.compile('^<ipython-input-([0-9]+)-.*>$')
 
 
-def get_local_reprs(frame, watch=(), custom_repr=(), max_length=None):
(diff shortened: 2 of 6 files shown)


cool-RR:
This looks excellent to me.

Two small notes:

1. `[False, True, ]` should be `(False, True)`
2. Where you're using `except NotImplementedError` you should use `pytest.raises`.

cool-RR:
@alexmojaki if you feel like taking a look, feel free, if you're busy that's cool.

cool-RR:
Wait, another note. Looking at the normalized output, I don't like how the code is now aligned with the variable changes. It's confusing. So when there's no datetime, instead of `''`, please use `' ' * 15`. That way the alignment is preserved. I guess you'll need to change the tests a bit too. Please do all these changes in separate commits than what you've submitted so far, so I could easily compare them.

cool-RR:
Good, the only thing that's left is that `[False, True, ]` should be `(False, True)`

itraviv:
i swear it was there! 😒

cool-RR:
Can you change it from `(True, False, )` to `(True, False)`?

itraviv:
> 
> 
> Can you change it from `(True, False, )` to `(True, False)`?

sure

cool-RR:
@alexmojaki Let me know whether you have time to give it a look before I merge.  (I'll squash Itamar's commits.)

alexmojaki:
Go ahead.

cool-RR:
I squashed these commits separately and made a release, adding Itamar to the authors file. Congrats on your first contribution!

itraviv:
thanks!

PR #174, opened 2019-12-14
add podcast link to readme

zachvalenta:


cool-RR:
I appreciate you adding this, but my voice came out difficult to hear on that podcast, and it was my fault because I was using a cheap headset and probably not even talking into the mic properly. So I'm not so keen on linking to it, sorry.

PR #178, opened 2020-04-15
Add elapsed time and elapsed_time format option.

iory:
Thank you for your awesome debug tool!

This PR added a printing total elapsed time and elapsed_time format option.
The test code is as follows.

```
import argparse
import time

import pysnooper


def test1(elapsed_time=False):
    with pysnooper.snoop(elapsed_time=elapsed_time):
        time.sleep(1.0)


def test2(elapsed_time=False):
    a = 1
    b = 2
    c = 3
    with pysnooper.snoop(elapsed_time=elapsed_time):
        tmp = a + b
        tmp2 = a - b


test1(elapsed_time=False)
print()
test1(elapsed_time=True)
print()
test2(elapsed_time=False)
print()
test2(elapsed_time=True)
```
The result is as follows.
```
Source path:... <ipython-input-14-f06356abdfce>
New var:....... elapsed_time = False
22:14:11.952148 line         9         time.sleep(1.0)
Total elapsed time: 00:00:01.002691

Source path:... <ipython-input-14-f06356abdfce>
New var:....... elapsed_time = True
00:00:00.000007 line         9         time.sleep(1.0)
Total elapsed time: 00:00:01.003509

Source path:... <ipython-input-14-f06356abdfce>
New var:....... elapsed_time = False
New var:....... a = 1
New var:....... b = 2
New var:....... c = 3
22:14:13.959043 line        17         tmp = a + b
New var:....... tmp = 3
22:14:13.959235 line        18         tmp2 = a - b
Total elapsed time: 00:00:00.000334

Source path:... <ipython-input-14-f06356abdfce>
New var:....... elapsed_time = True
New var:....... a = 1
New var:....... b = 2
New var:....... c = 3
00:00:00.000011 line        17         tmp = a + b
New var:....... tmp = 3
00:00:00.000295 line        18         tmp2 = a - b
Total elapsed time: 00:00:00.000409
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,18 +72,18 @@ Or if you don't want to trace an entire function, you can wrap the relevant part
 import pysnooper
 import random
 
-def foo():
+def foo(relative_time=False):
     lst = []
     for i in range(10):
         lst.append(random.randrange(1, 1000))
 
-    with pysnooper.snoop():
+    with pysnooper.snoop(relative_time=relative_time):
         lower = min(lst)
         upper = max(lst)
         mid = (lower + upper) / 2
         print(lower, mid, upper)
 
-foo()
+foo(False)
 ```
 
 which outputs something like:
@@ -99,6 +99,27 @@ New var:....... upper = 832
 74 453.0 832
 New var:....... mid = 453.0
 09:37:35.882486 line        13         print(lower, mid, upper)
+Total elapsed time: 00:00:00.000344
+```
+
+If `relative_time` is `True`, print time format will be relative.
+
+```python
+import pysnooper
+import random
+
+def foo(relative_time=False):
+    lst = []
+    for i in range(10):
+        lst.append(random.randrange(1, 1000))
+
+    with pysnooper.snoop(relative_time=relative_time):
+        lower = min(lst)
+        upper = max(lst)
+        mid = (lower + upper) / 2
+        print(lower, mid, upper)
+
+foo(True)
 ```
 
 # Features #
--- a/README.md
+++ b/README.md
@@ -72,12 +72,12 @@ Or if you don't want to trace an entire function, you can wrap the relevant part
 import pysnooper
 import random
 
-def foo(relative_time=False):
+def foo(elapsed_time=False):
     lst = []
     for i in range(10):
         lst.append(random.randrange(1, 1000))
 
-    with pysnooper.snoop(relative_time=relative_time):
+    with pysnooper.snoop(elapsed_time=elapsed_time):
         lower = min(lst)
         upper = max(lst)
         mid = (lower + upper) / 2
@@ -102,18 +102,18 @@ New var:....... mid = 453.0
 Total elapsed time: 00:00:00.000344
 ```
 
-If `relative_time` is `True`, print time format will be relative.
+If `elapsed_time` is `True`, print elapsed time format.
 
 ```python
 import pysnooper
 import random
 
-def foo(relative_time=False):
+def foo(elapsed_time=False):
     lst = []
     for i in range(10):
         lst.append(random.randrange(1, 1000))
 
-    with pysnooper.snoop(relative_time=relative_time):
+    with pysnooper.snoop(elapsed_time=elapsed_time):
         lower = min(lst)
         upper = max(lst)
         mid = (lower + upper) / 2
--- a/pysnooper/pycompat.py
+++ b/pysnooper/pycompat.py
@@ -80,3 +80,9 @@ else:
         return result
 
 
+def timedelta_isoformat(timedelta, timespec='microseconds'):
+    assert isinstance(timedelta, datetime_module.timedelta)
+    if timespec != 'microseconds':
+        raise NotImplementedError
+    time = (datetime_module.datetime.min + timedelta).time()
+    return time_isoformat(time, timespec)
--- a/pysnooper/pycompat.py
+++ b/pysnooper/pycompat.py
@@ -90,9 +90,6 @@ else:
         return datetime_module.time(hour, minute, second, microsecond)
 
 
-def timedelta_isoformat(timedelta, timespec='microseconds'):
-    assert isinstance(timedelta, datetime_module.timedelta)
-    if timespec != 'microseconds':
-        raise NotImplementedError
+def timedelta_isoformat(timedelta):
     time = (datetime_module.datetime.min + timedelta).time()
-    return time_isoformat(time, timespec)
+    return time_isoformat(time)
--- a/pysnooper/pycompat.py
+++ b/pysnooper/pycompat.py
@@ -90,6 +90,6 @@ else:
         return datetime_module.time(hour, minute, second, microsecond)
 
 
-def timedelta_isoformat(timedelta):
+def timedelta_isoformat(timedelta, timespec='microseconds'):
     time = (datetime_module.datetime.min + timedelta).time()
-    return time_isoformat(time)
+    return time_isoformat(time, timespec)
--- a/pysnooper/pycompat.py
+++ b/pysnooper/pycompat.py
@@ -80,6 +80,16 @@ else:
         return result
 
 
+if sys.version_info[:2] >= (3, 7):
+    time_fromisoformat = datetime_module.time.fromisoformat
+else:
+    def time_fromisoformat(isoformat_str):
+        hour, minute, second, microsecond = map(
+            int,
+            isoformat_str.replace('.', ':').split(':'))
+        return datetime_module.time(hour, minute, second, microsecond)
+
+
 def timedelta_isoformat(timedelta, timespec='microseconds'):
     assert isinstance(timedelta, datetime_module.timedelta)
     if timespec != 'microseconds':
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -217,7 +217,7 @@ class Tracer:
              for v in utils.ensure_tuple(watch_explode)
         ]
         self.frame_to_local_reprs = {}
-        self.start_times = {}
+        self.start_times = []
         self.depth = depth
         self.prefix = prefix
         self.thread_info = thread_info
@@ -303,7 +303,7 @@ class Tracer:
             'original_trace_functions', []
         )
         stack.append(sys.gettrace())
-        self.start_times[id(calling_frame)] = datetime_module.datetime.now()
+        self.start_times.append(datetime_module.datetime.now())
         sys.settrace(self.trace)
 
     def __exit__(self, exc_type, exc_value, exc_traceback):
@@ -315,7 +315,7 @@ class Tracer:
         self.target_frames.discard(calling_frame)
         self.frame_to_local_reprs.pop(calling_frame, None)
 
-        start_time = self.start_times.pop(id(calling_frame))
+        start_time = self.start_times.pop(-1)
         duration = datetime_module.datetime.now() - start_time
         now_string = pycompat.timedelta_isoformat(duration, timespec='microseconds')
         indent = ' ' * 4 * (thread_global.depth + 1)
@@ -332,7 +332,6 @@ class Tracer:
         return thread_info.ljust(self.thread_info_padding)
 
     def trace(self, frame, event, arg):
-
         ### Checking whether we should trace this line: #######################
         #                                                                     #
         # We should trace this line either if it's in the decorated function,
@@ -367,9 +366,7 @@ class Tracer:
         ### Finished checking whether we should trace this line. ##############
 
         if self.elapsed_time:
-            calling_frame = frame.f_back
-            duration = datetime_module.datetime.now() - self.start_times[
-                id(calling_frame)]
+            duration = datetime_module.datetime.now() - self.start_times[-1]
             now_string = pycompat.timedelta_isoformat(
                 duration, timespec='microseconds') if not self.normalize else ' ' * 15
         else:
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -310,9 +310,8 @@ class Tracer:
         self.target_frames.discard(calling_frame)
         self.frame_to_local_reprs.pop(calling_frame, None)
 
-        now = (datetime_module.datetime.min + (
-            datetime_module.datetime.now() - self.start_time)).time()
-        now_string = pycompat.time_isoformat(now, timespec='microseconds')
+        duration = datetime_module.datetime.now() - self.start_time
+        now_string = pycompat.timedelta_isoformat(duration, timespec='microseconds')
         self.write('Total elapsed time: {now_string}'.format(**locals()))
 
     def _is_internal_frame(self, frame):
@@ -360,11 +359,12 @@ class Tracer:
         ### Finished checking whether we should trace this line. ##############
 
         if self.relative_time:
-            now = (datetime_module.datetime.min +
-                   (datetime_module.datetime.now() - self.start_time)).time()
+            duration = datetime_module.datetime.now() - self.start_time
+            now_string = pycompat.timedelta_isoformat(
+                duration, timespec='microseconds') if not self.normalize else ' ' * 15
         else:
             now = datetime_module.datetime.now().time()
-        now_string = pycompat.time_isoformat(now, timespec='microseconds') if not self.normalize else ' ' * 15
+            now_string = pycompat.time_isoformat(now, timespec='microseconds') if not self.normalize else ' ' * 15
         line_no = frame.f_lineno
         source_path, source = get_path_and_source_from_frame(frame)
         source_path = source_path if not self.normalize else os.path.basename(source_path)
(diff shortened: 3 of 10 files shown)


cool-RR:
Looks cool! I'll review it when I have time.

cool-RR:
Looks good.

1. In the code you added to `__exit__`, you should do `duration = datetime_module.datetime.now() - self.start_time`. Then `duration` is a `timedelta` object. You should create a `timedelta_isoformat` function in the `pycompat` module, where you could put the hack you've used in lines 313-314. Make sure there's a bit of testing for that hack.
2. Add tests for both the features you added.
3. Add documentation for `relative_time` similarly to the other flags.
4. Modify the example in the readme to show the duration like your new code shows.
5. Lines 362-366 are also hacky, but I'll fix them after you finish, so don't worry about that.

alexmojaki:
There's just one start time being stored in the tracer. How does this handle the same tracer tracking multiple calls at the same time? You probably need a mapping from frames to times.

alexmojaki:
There's just one start time being stored in the tracer. How does this handle the same tracer tracking multiple calls at the same time? You probably need a mapping from frames to times.

cool-RR:
@alexmojaki Nice catch, I didn't think of that. Let's see what @iory comes up with. 

@iory I'm happy to see these commits. When you're ready for me to review, let me know.

iory:
@alexmojaki  
Thank you for your comment. I would like to cope with that case.
Could you let me know the test case?

@cool-RR 
I've fixed it based on your comments. Please review it.

alexmojaki:
See for example reusing a tracer in https://github.com/cool-RR/PySnooper/blob/cba23d18a8809aef304718ddfec32d6249e12b34/tests/test_pysnooper.py#L900

Or have a test case with a recursive call.

Also try to test that the outputted time makes sense, e.g. put a sleep(0.1) somewhere and make sure the elapsed time is greater than that.

Are these tests passing? I don't see any normalising of the output for `assert_sample_output`.

cool-RR:
@bittner Didn't we use to have tests running automatically on every PR? I thought we did but I'm not seeing them running now. Am I misremembering?

cool-RR:
@iory I reviewed the code. There are a few problems, but I'll fix them when your PR is ready so don't worry about these. Address the 3 points that Alex raised and I think we'll be done.

iory:
@alexmojaki 
Thank you for your reviews!
I modified the code and coped with the recursive call case. (also I checked pytest passed.)
Could you check the code?

The Test case's result is as follows

```
import time

import pysnooper

snoop = pysnooper.snoop(elapsed_time=True, depth=2)


def foo(x, depth):
    if x == 0:
        bar1(x, depth)
        qux()
        return

    with snoop:
        # There should be line entries for these three lines,
        # no line entries for anything else in this function,
        # but calls to all bar functions should be traced
        foo(x - 1, depth + 1)
        bar2(x, depth)
        qux()
    int(4)
    bar3(9, depth)
    return x


@snoop
def bar1(_x, depth):
    qux()


@snoop
def bar2(_x, depth):
    qux()


@snoop
def bar3(_x, depth):
    qux()


def qux():
    time.sleep(1.0)
    return 9  # not traced, mustn't show up


result = foo(2, 1)

Source path:... <ipython-input-1-5287e53df31e>
New var:....... x = 2
New var:....... depth = 1
00:00:00.000019 line        18         foo(x - 1, depth + 1)
Starting var:.. x = 1
Starting var:.. depth = 2
00:00:00.000240 call         8 def foo(x, depth):
00:00:00.000307 line         9     if x == 0:
00:00:00.000367 line        14     with snoop:
00:00:00.000015 line        18         foo(x - 1, depth + 1)
    Starting var:.. x = 0
    Starting var:.. depth = 3
    00:00:00.000112 call         8 def foo(x, depth):
    00:00:00.000221 line         9     if x == 0:
    00:00:00.000290 line        10         bar1(x, depth)
        Starting var:.. _x = 0
        Starting var:.. depth = 3
        00:00:00.000011 call        27 def bar1(_x, depth):
        00:00:00.000119 line        28     qux()
            00:00:00.000188 call        41 def qux():
            00:00:00.000249 line        42     time.sleep(1.0)

            00:00:01.004509 line        43     return 9  # not traced, mustn't show up
            00:00:01.004653 return      43     return 9  # not traced, mustn't show up
            Return value:.. 9
        00:00:01.004749 return      28     qux()
        Return value:.. None
        Total elapsed time: 00:00:01.004881
    00:00:01.005316 line        11         qux()
    00:00:02.008273 line        12         return
    00:00:02.008526 return      12         return
    Return value:.. None
00:00:02.008684 line        19         bar2(x, depth)
    Starting var:.. _x = 1
    Starting var:.. depth = 2
    00:00:00.000010 call        32 def bar2(_x, depth):
    00:00:00.000160 line        33     qux()
        00:00:00.000247 call        41 def qux():
        00:00:00.000303 line        42     time.sleep(1.0)
        00:00:01.005015 line        43     return 9  # not traced, mustn't show up
        00:00:01.005159 return      43     return 9  # not traced, mustn't show up
        Return value:.. 9
    00:00:01.005244 return      33     qux()
    Return value:.. None
    Total elapsed time: 00:00:01.005352
00:00:03.014215 line        20         qux()
    00:00:03.014284 call        41 def qux():
    00:00:03.014337 line        42     time.sleep(1.0)
    00:00:04.015809 line        43     return 9  # not traced, mustn't show up
    00:00:04.015960 return      43     return 9  # not traced, mustn't show up
    Return value:.. 9
    Total elapsed time: 00:00:04.016078
New var:....... x = 1
New var:....... depth = 2
00:00:04.016723 line        21     int(4)
00:00:04.016810 line        22     bar3(9, depth)
    Starting var:.. _x = 9
    Starting var:.. depth = 2
    00:00:00.000007 call        37 def bar3(_x, depth):
    00:00:00.000085 line        38     qux()
        00:00:00.000232 call        41 def qux():
        00:00:00.000365 line        42     time.sleep(1.0)
        00:00:01.003332 line        43     return 9  # not traced, mustn't show up
        00:00:01.003537 return      43     return 9  # not traced, mustn't show up
        Return value:.. 9
    00:00:01.003692 return      38     qux()
    Return value:.. None
    Total elapsed time: 00:00:01.003856
00:00:05.020806 line        23     return x
00:00:05.020910 return      23     return x
Return value:.. 1
00:00:05.020978 line        19         bar2(x, depth)
Starting var:.. _x = 2
Starting var:.. depth = 1
00:00:00.000007 call        32 def bar2(_x, depth):
00:00:00.000084 line        33     qux()
    00:00:00.000143 call        41 def qux():
    00:00:00.000231 line        42     time.sleep(1.0)
    00:00:01.002687 line        43     return 9  # not traced, mustn't show up
    00:00:01.002878 return      43     return 9  # not traced, mustn't show up
    Return value:.. 9
00:00:01.002994 return      33     qux()
Return value:.. None
Total elapsed time: 00:00:01.003124
00:00:06.024215 line        20         qux()
00:00:06.024290 call        41 def qux():
00:00:06.024346 line        42     time.sleep(1.0)
00:00:07.025803 line        43     return 9  # not traced, mustn't show up
00:00:07.026055 return      43     return 9  # not traced, mustn't show up
Return value:.. 9
Total elapsed time: 00:00:07.026177
Starting var:.. _x = 9
Starting var:.. depth = 1
00:00:00.000009 call        37 def bar3(_x, depth):
00:00:00.000104 line        38     qux()
    00:00:00.000162 call        41 def qux():
    00:00:00.000208 line        42     time.sleep(1.0)
    00:00:01.001732 line        43     return 9  # not traced, mustn't show up
    00:00:01.001922 return      43     return 9  # not traced, mustn't show up
    Return value:.. 9
00:00:01.002199 return      38     qux()
Return value:.. None
Total elapsed time: 00:00:01.002651
```

README.md
```diff
@@ -72,18 +72,18 @@ Or if you don't want to trace an entire function, you can wrap the relevant part
...
 import random
 
-def foo():
+def foo(elapsed_time=False):
     lst = []
     for i in range(10):
         lst.append(random.randrange(1, 1000))
 
-    with pysnooper.snoop():
+    with pysnooper.snoop(elapsed_time=elapsed_time):
```
alexmojaki, line 80: `elapsed_time` should not be mentioned here, this section is about `with`.

cool-RR: That's correct, but I'll fix the readme after we approve this PR, so you can ignore that in your review.

README.md
```diff
@@ -99,6 +99,27 @@ New var:....... upper = 832
...
+```python
+import pysnooper
+import random
+
+def foo(elapsed_time=False):
+    lst = []
+    for i in range(10):
+        lst.append(random.randrange(1, 1000))
+
+    with pysnooper.snoop(elapsed_time=elapsed_time):
```
alexmojaki, line 116: Just write `with pysnooper.snoop(elapsed_time=True):` and keep the example simple. No good reason to pass arguments around.

This particular code was used to demonstrate the advantage of `with` - in particular, nothing from the for loop above appears in the output. Use different code that is well suited for demonstrating the usefulness of `elapsed_time`. It can be much shorter and should probably have some sleeps in it.

Show the output of `elapsed_time=True`, and explain what the difference is. There will always be a line at the end `Total elapsed time:` even with `elapsed_time=False` - that is already confusing and potentially even undesirable. If that's not going to change, at least make it clear in documentation. People will easily miss that the numbers in the time column have changed with `elapsed_time=True`.

All of this should be much lower down anyway, probably under advanced usage. This section of the README is for the basics.

pysnooper/tracer.py
```diff
@@ -198,10 +198,15 @@ class Tracer:
 
     You can also use `max_variable_length=None` to never truncate them.
 
+    Print time in elapsed time format::
```
alexmojaki, line 201: "elapsed time format" is a confusing description. Explain more here.

pysnooper/tracer.py
```diff
@@ -198,10 +198,15 @@ class Tracer:
...
+
+        @pysnooper.snoop(elapsed_time=True)
+
     '''
     def __init__(self, output=None, watch=(), watch_explode=(), depth=1,
                  prefix='', overwrite=False, thread_info=False, custom_repr=(),
-                 max_variable_length=100, normalize=False):
+                 max_variable_length=100, normalize=False,
```
alexmojaki, line 208: I think it's high time that this signature had one parameter per line, @cool-RR ?

pysnooper/tracer.py
```diff
@@ -212,6 +217,7 @@ def __init__(self, output=None, watch=(), watch_explode=(), depth=1,
              for v in utils.ensure_tuple(watch_explode)
         ]
         self.frame_to_local_reprs = {}
+        self.start_times = []
```
alexmojaki, line 220: A stack is not good enough, this will not handle multiple threads. As I said before, you need a mapping from frames to times, similar to `frame_to_local_reprs`.

README.md
cool-RR: I'll fix this later

pysnooper/tracer.py
cool-RR: I'll fix this later.

pysnooper/tracer.py
cool-RR: Alex is probably right here.

pysnooper/pycompat.py
```diff
@@ -80,3 +80,19 @@ def time_isoformat(time, timespec='microseconds'):
...
 
 
+if sys.version_info[:2] >= (3, 7):
+    time_fromisoformat = datetime_module.time.fromisoformat
+else:
+    def time_fromisoformat(isoformat_str):
+        hour, minute, second, microsecond = map(
+            int,
+            isoformat_str.replace('.', ':').split(':'))
+        return datetime_module.time(hour, minute, second, microsecond)
```
alexmojaki, line 90: @cool-RR what do you think of

```suggestion
        parts = map(int, isoformat_str.replace('.', ':').split(':'))
        return datetime_module.time(*parts)
```

pysnooper/tracer.py
cool-RR: I don't like those too much. They take up too much vertical space.

cool-RR: I'll fix the line wrapping later.

pysnooper/pycompat.py
```diff
@@ -80,3 +80,19 @@ def time_isoformat(time, timespec='microseconds'):
...
+
+
+def timedelta_isoformat(timedelta, timespec='microseconds'):
+    assert isinstance(timedelta, datetime_module.timedelta)
+    if timespec != 'microseconds':
+        raise NotImplementedError
```
alexmojaki, line 96:
```suggestion
def timedelta_isoformat(timedelta):
```

There's no need to handle `timespec` as this doesn't need to be compatible with an existing method. And the line below will fail nicely if a non-timedelta is passed.

pysnooper/tracer.py
```diff
@@ -307,6 +315,13 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
         self.target_frames.discard(calling_frame)
         self.frame_to_local_reprs.pop(calling_frame, None)
 
+        start_time = self.start_times.pop(-1)
+        duration = datetime_module.datetime.now() - start_time
+        now_string = pycompat.timedelta_isoformat(duration, timespec='microseconds')
```
alexmojaki, line 320:
```suggestion
        now_string = pycompat.timedelta_isoformat(duration)
```

No point in passing the default value.

pysnooper/pycompat.py
cool-RR: Let's leave this for later. I want to get the pr merged before the author gets too frustrated, and then improve it later. This also goes for all your other comments, except the frame dict.

tests/samples/recursion.py
```diff
@@ -18,45 +18,49 @@ def main():
 expected_output = '''
 Source path:... Whatever
 Starting var:.. x = 4
-20:28:17.875295 call         5 def factorial(x):
-20:28:17.875509 line         6     if x <= 1:
-20:28:17.875550 line         8     return mul(x, factorial(x - 1))
+09:31:32.691599 call         5 def factorial(x):
```
alexmojaki, line 21: This is not ideal, it ruins the git history.

pysnooper/tracer.py
iory: Delete default value.

tests/samples/recursion.py
iory: OK. I'll modify and rebase it.

alexmojaki:
I see you've tested that the time after "Total elapsed time:" is correct, but there's no testing of the time that appears in every line.

pysnooper/pycompat.py
iory: Why do you suggest deleting the assertion of checking datetime_module.timedelta?

alexmojaki: It's just noise in the code, and as I said the type is enforced by the rest of the code anyway. If you call `timedelta_isoformat(3)`, without the assertion you immediately get an error `TypeError: unsupported operand type(s) for +: 'datetime.datetime' and 'int'`. With the assertion, you get:

```
    assert isinstance(timedelta, datetime_module.timedelta)
AssertionError
```

which is actually less informative.

iory: I see. I agree with your idea.

alexmojaki:
By the way, what's the use case for all this? Why not use a profiling tool specially designed for the task? Have you tried tools like:

- https://github.com/csurfer/pyheat
- https://github.com/uber-archive/pyflame or https://github.com/baverman/flameprof/
- https://docs.python.org/3.8/library/profile.html

iory:
> what's the use case for all this? 

I want to know what the value is and how long it takes while debugging.
Pysnooper is a poor man's debugger, you know?
Is this feature not suitable for the pysnooper concept?

alexmojaki:
Even assuming that it's suitable, it's good to think carefully about what exactly we want to achieve and then design the user interface around that.

cool-RR:
I think that both features are good. Yes, there are dedicated profilers, but since PySnooper is already there and has the information, and it's dumping a big chunk of text anyway, it's helpful to include it.

cool-RR:
I suggest that we freeze the review. I'll merge later today and add my fixes, and then both of you could reopen any discussions for another pr.

alexmojaki:
Still need to track times per frame.

cool-RR:
Right, so let's limit the discussion to this.

iory:
Thank you for yours discussion.
I modified `start_frames` like `frame_to_local_reprs`.

bittner:
> @bittner Didn't we use to have tests running automatically on every PR? I thought we did but I'm not seeing them running now. Am I misremembering?

The builds do run, but the integration with GitHub seems broken. Hence the green tick doesn't show next to commits and new PRs.

You may want to check the repository settings. It's probably that you withdrew permission for Travis CI unconsciously or by mistake.

cool-RR:
@bittner Thanks, I fixed that now.

@iory , I fixed some of the issues, merged the commits and made a new release. Congrats on your contribution!

@iory @alexmojaki Feel free to look into my commit and see the changes I made. If any of them could be a problem, or you have a possible improvement, please open an issue/PR.

iory:
Thank you!

PR #196, opened 2020-09-13
fix #195

### LINKED ISSUE #195: '_thread._local' object has no attribute 'depth' raised by snooper if there's exception upon calling snooped func
```
import pysnooper
@pysnooper.snoop()
def f(x):
    pass
```
Calling f without x
```
----> 1 f()
```

results in:
```
.../pysnooper/tracer.py in simple_wrapper(*args, **kwargs)
    261         def simple_wrapper(*args, **kwargs):
    262             with self:
--> 263                 return function(*args, **kwargs)
    264
    265         @functools.wraps(function)

.../pysnooper/tracer.py in __exit__(self, exc_type, exc_value, exc_traceback)
    320         duration = datetime_module.datetime.now() - start_time
    321         elapsed_time_string = pycompat.timedelta_format(duration)
--> 322         indent = ' ' * 4 * (thread_global.depth + 1)
    323         self.write(
    324             '{indent}Elapsed time: {elapsed_time_string}'.format(**locals())

AttributeError: '_thread._local' object has no attribute 'depth'
```
The reason is that we initialize `thread_global` with `depth`here, while it's always used on `__exit__` here:https://github.com/cool-RR/PySnooper/blob/master/pysnooper/tracer.py#L322. In case we returned before the thread_global was initialized with depth, (i.e. https://github.com/cool-RR/PySnooper/blob/master/pysnooper/tracer.py#L351), the thread_global won't have the depth attr.

yaelmi3:
Fix '_thread._local' object has no attribute 'depth' raised by snooper if there's exception upon calling snooped func by initializing thread_global dict with depth when entering the trace() func

### DIFF
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -293,6 +293,7 @@ class Tracer:
     def __enter__(self):
         if DISABLED:
             return
+        thread_global.__dict__.setdefault('depth', -1)
         calling_frame = inspect.currentframe().f_back
         if not self._is_internal_frame(calling_frame):
             calling_frame.f_trace = self.trace
@@ -362,7 +363,6 @@ class Tracer:
                 else:
                     return None
 
-        thread_global.__dict__.setdefault('depth', -1)
         if event == 'call':
             thread_global.depth += 1
         indent = ' ' * 4 * thread_global.depth
--- a/tests/test_pysnooper.py
+++ b/tests/test_pysnooper.py
@@ -1894,5 +1894,10 @@ def test_exception():
     )
 
 
+def test_exception_on_entry():
+    @pysnooper.snoop()
+    def f(x):
+        pass
 
-
+    with pytest.raises(TypeError):
+        f()


cool-RR:
Good job.

@yaelmi3 : As you can see, your test is failing on Python 2 because the exception message is different. Please change your test to not check the exception message at all. Confirming it's a `TypeError` is sufficient.

@alexmojaki : It looks like Yael took the `thread_global.__dict__.setdefault('depth', -1)` line to the top of `trace`, which solved the problem. Do you think it'll make more sense to put it in `__enter__` instead? Any other insights would be appreciated.

cool-RR:
@yaelmi3 Great, now waiting for Alex's input.

alexmojaki:
> Do you think it'll make more sense to put it in `__enter__` instead?

I guess so. Should be a bit faster.

cool-RR:
@yaelmi3 mind moving that line?

yaelmi3:
Done

cool-RR:
@yaelmi3 Congrats on your first contribution. I added you to the list of authors and made a new release with your bugfix :+1:

PR #212, opened 2021-02-26
Remove ad

cool-RR:


cool-RR:
@bittner Looks like the CI is broken :cry: 

It's possible that this has something to do with Travis's strategy change. Or maybe it's just a bug, I don't know. Do you have time to take a look?

bittner:
The `six` module can't be found. It was bundled in commit 4c0b64fd0bc14ff2 and removed in a73ea14f585ab2, but is still used in [tests/mini_toolbox/pathlib.py](https://github.com/cool-RR/PySnooper/blob/master/tests/mini_toolbox/pathlib.py#L13).

At first sight, this looks like a bug.

cool-RR:
@bittner Fixed :)

Issue #218, opened 2021-05-11
Some tests are failing with Python 3.10 beta 1


frenzymadness:
Hello.

It seems that some parts of PySnooper produce different results in Python 3.10 beta 1 and therefore 9 tests fail:

```
================================== short test summary info ===================================
FAILED tests/test_pysnooper.py::test_relative_time - tests.utils.OutputFailure: 
FAILED tests/test_pysnooper.py::test_with_block[True] - tests.utils.OutputFailure: 
FAILED tests/test_pysnooper.py::test_with_block[False] - tests.utils.OutputFailure: 
FAILED tests/test_pysnooper.py::test_with_block_depth[True] - tests.utils.OutputFailure: 
FAILED tests/test_pysnooper.py::test_with_block_depth[False] - tests.utils.OutputFailure: 
FAILED tests/test_pysnooper.py::test_cellvars[True] - tests.utils.OutputFailure: 
FAILED tests/test_pysnooper.py::test_cellvars[False] - tests.utils.OutputFailure: 
FAILED tests/test_pysnooper.py::test_var_order[True] - tests.utils.OutputFailure: 
FAILED tests/test_pysnooper.py::test_var_order[False] - tests.utils.OutputFailure: 
========================= 9 failed, 65 passed, 21 warnings in 1.47s ==========================
```
But, I've tried to reproduce the issue manually and I have to admit that I see no real difference between the output of a simple script run via Python 3.9 and 3.10. Might it be caused by something internal in tests?

hroncok:
Do you run beta 1? Because it used to work with alpha 7.

frenzymadness:
Yes. I guess that the switch from alpha 7 to beta 1 is the reason why I did not get this report earlier.

alexmojaki:
The tests for my fork `snoop` make it a little easier to see the difference: https://github.com/alexmojaki/snoop/commit/f83c0ae6767f3873c03f5d1c4a0af7fc96ae3219

Looks like the main difference is that the `with` line gets revisited on exiting. I saw something recently about sys.settrace visiting lines more thoroughly, that's probably why this is happening.

frenzymadness:
You are right, when I use a with block, there is one more line in Python 3.10:

Code
```
import pysnooper

@pysnooper.snoop()
def f(foo):
    with open("cool", mode="w") as f:
        f.write(foo)

f("bar")
```
Python 3.9:
```
[...]

PR #219, opened 2021-05-18
Python 3.10 compatibility

frenzymadness:
Fixes: #218 

This is the minimum I had to implement to make the tests compatible with Python 3.10. I also did some improvements in configs for CI and tests.

I'm not sure what is your approach to flake8 and pylint because there are already many errors and it's hard to find the ones possible caused by me here. Also, it's kinda verbose to use the names for keyword arguments instead of `**kwargs` for the optional ones but I've followed the current codebase.

All unit tests are passing locally so we'll see what CI thinks about it.


cool-RR:
This looks good to me. @alexmojaki Do you want to take a look? 

A few notes:
1. Anywhere you use `(3,10)`, space it out: `(3, 10)`
2. I don't like that the `expected_entries` variable gets replaced. Can you remove the assignment `expected_entries = filtered_expected_entries` and just iterate over `filtered_expected_entries` in the `for` loop?
3. The CI is failing for some reason, but I guess you noticed that.

frenzymadness:
Thanks for the feedback. I've tried to improve the logic to be more readable. See the latest fixup commit.

The CI fails because Travis has not the latest Python 3.10 beta 1 yet, it's still on alpha 5. If you want to, I can add the 3.10 job into `allow_failures` and we can remove it from there once Travis updates its Python.

cool-RR:
Great job. Yes, using `allow_failures` sounds good to me.

alexmojaki:
No comments from me because I don't like the 'Entry' style tests anyway.

cool-RR:
Hmm, any idea why Travis still fails?

frenzymadness:
The configuration is kinda different when handled by the tox-travis, let me find the right way.

cool-RR:
Okay, Travis is green. Are we good to merge? (After the merge I'll do a version bump and make a release.)

frenzymadness:
I think we are. Thank you!

cool-RR:
Done and release. Thank you.

cool-RR:
Right, done.

PR #220, opened 2021-07-26
fix an indent bugs and add some new functions

Undefeated-man:
1) make it available to identify which file the called function belongs to when our traceback depth is bigger than 1 -- automatically add the name of the file before printing the variables

2) adding "\n" before each "newish_string" and "Source path: xxx"  to make the output nicer and easier to look

3) fixing an indent mistake at lines 363-364 in "tracer.py"

for example, this is my demo when depth=2:

### DIFF
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -360,8 +360,8 @@ class Tracer:
                         return None
                     elif _frame_candidate.f_code in self.target_codes or _frame_candidate in self.target_frames:
                         break
-                    else:
-                        return None
+                else:
+                    return None
 
         if event == 'call':
             thread_global.depth += 1
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -360,8 +360,8 @@ class Tracer:
                         return None
                     elif _frame_candidate.f_code in self.target_codes or _frame_candidate in self.target_frames:
                         break
-                else:
-                    return None
+                    else:
+                        return None
 
         if event == 'call':
             thread_global.depth += 1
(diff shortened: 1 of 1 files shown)


pysnooper/tracer.py
```diff
@@ -360,8 +360,8 @@ def trace(self, frame, event, arg):
                         return None
                     elif _frame_candidate.f_code in self.target_codes or _frame_candidate in self.target_frames:
                         break
-                else:
```
cool-RR, line 363: Reading the code, I don't understand why your interpretation is better. Can you explain?

cool-RR (commented):
I don't really see changes 1 and 2 making the output clearer, sorry.

pysnooper/tracer.py
Undefeated-man: Thanks for your patient reply!😁 And may I explain my idea clearer cause it seems that I didn't interpret it clear and even made a mistake at the (3) point~

For the (1) interpretation, I mean when our project is really huge, we often import many things from other files, and maybe some functions have the same names but are in different models. When our trackback depth is bigger than 1, sometimes we may not easy to know actually which function we are recording. Adding the name of ".py" files at the beginning of the line can help us identify actually which function we are tracking back at that specific recording line.

And for the (3), I'm sorry that I made a mistake that mistook "for...else..." for an indent mistake. Quite sorry for this one.

For the (2) one, I think our output log is kind of messy right now, if we add some line breaks after every time we modified variables, and forming them as many tiny blocks of records may help a lot.

cool-RR: The newlines may help, but they also might confuse when we have code that changes between a lot of files, so I don't want to add this. 
If you want to make PySnooper output easier to look at, you could make a program that creates HTML out of that output, and then you could get very fancy with the formatting, folding, etc.

cool-RR:
Thanks anyway for trying to contribute.

PR #226, opened 2022-03-23
Add support for Ansible zipped source files

luenk:
# Ansible zipped source support

## Scope
It would be great to support Ansible zipped source files to make Ansible module debugging available in PySnooper

## Problem
Ansible is using zipped source files on the remote node. 
To get access to the source file itself you need to unzip the file during runtime.
The source file will be deleted after execution.

```
Source path:... /tmp/ansible_my_module_payload_xyz1234/ansible_my_module_payload.zip/ansible/modules/my_module.py
Modified var:.. result = {'changed': True, 'original_message': 'hello', 'message': 'goodbye'}
09:25:25.942249 line       127 SOURCE IS UNAVAILABLE
09:25:25.942279 line       132 SOURCE IS UNAVAILABLE
09:25:25.942303 line       133 SOURCE IS UNAVAILABLE
```

## Solution
I added a condition to tracer.py like the condition for ipython to unzip the source file and get read access. 

```
Source path:... /tmp/ansible_my_module_payload_xyz1234/ansible_my_module_payload.zip/ansible/modules/my_module.py
Modified var:.. result = {'changed': True, 'original_message': 'hello', 'message': 'goodbye'}
09:26:58.498181 line       127     if module.params['name'] == 'fail me':
09:26:58.498211 line       132     if module.params['debug']:
09:26:58.498235 line       133         result['debug'] = pydebug.getvalue()
```

### DIFF
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -20,6 +20,7 @@ if pycompat.PY2:
 
 
 ipython_filename_pattern = re.compile('^<ipython-input-([0-9]+)-.*>$')
+ansible_filename_pattern = re.compile('^(/.+zip)/(ansible/modules/.+)$')
 
 
 def get_local_reprs(frame, watch=(), custom_repr=(), max_length=None, normalize=False):
@@ -67,6 +68,7 @@ def get_path_and_source_from_frame(frame):
             source = source.splitlines()
     if source is None:
         ipython_filename_match = ipython_filename_pattern.match(file_name)
+        ansible_filename_match = ansible_filename_pattern.match(file_name)
         if ipython_filename_match:
             entry_number = int(ipython_filename_match.group(1))
             try:
@@ -75,6 +77,13 @@ def get_path_and_source_from_frame(frame):
                 ((_, _, source_chunk),) = ipython_shell.history_manager. \
                                   get_range(0, entry_number, entry_number + 1)
                 source = source_chunk.splitlines()
+            except Exception:
+                pass
+        elif ansible_filename_match:
+            try:
+                import zipfile
+                archive_file = zipfile.ZipFile(ansible_filename_match.group(1), 'r')
+                source = archive_file.read(ansible_filename_match.group(2)).splitlines()
             except Exception:
                 pass
         else:
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -20,7 +20,7 @@ if pycompat.PY2:
 
 
 ipython_filename_pattern = re.compile('^<ipython-input-([0-9]+)-.*>$')
-ansible_filename_pattern = re.compile(r'^(/.+\.zip)/(ansible/modules/.+\.py)$')
+ansible_filename_pattern = re.compile(r'^(.+\.zip)[/|\\](ansible[/|\\]modules[/|\\].+\.py)$')
 
 
 def get_local_reprs(frame, watch=(), custom_repr=(), max_length=None, normalize=False):
@@ -83,7 +83,7 @@ def get_path_and_source_from_frame(frame):
             try:
                 import zipfile
                 archive_file = zipfile.ZipFile(ansible_filename_match.group(1), 'r')
-                source = archive_file.read(ansible_filename_match.group(2)).splitlines()
+                source = archive_file.read(ansible_filename_match.group(2).replace('\\', '/')).splitlines()
             except Exception:
                 pass
         else:
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -20,7 +20,7 @@ if pycompat.PY2:
 
 
 ipython_filename_pattern = re.compile('^<ipython-input-([0-9]+)-.*>$')
-ansible_filename_pattern = re.compile('^(/.+zip)/(ansible/modules/.+)$')
+ansible_filename_pattern = re.compile(r'^(/.+\.zip)/(ansible/modules/.+\.py)$')
 
 
 def get_local_reprs(frame, watch=(), custom_repr=(), max_length=None, normalize=False):
--- a/tests/test_pysnooper.py
+++ b/tests/test_pysnooper.py
@@ -9,7 +9,6 @@ import types
 import os
 import sys
 import zipfile
-import importlib
 
 from pysnooper.utils import truncate
 import pytest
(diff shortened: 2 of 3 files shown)


pysnooper/tracer.py
```diff
@@ -20,6 +20,7 @@
 
 
 ipython_filename_pattern = re.compile('^<ipython-input-([0-9]+)-.*>$')
+ansible_filename_pattern = re.compile('^(/.+zip)/(ansible/modules/.+)$')
```
cool-RR, line 23: Please use a raw string here, and also fix the line before it to be a raw string. 

I want to be extra careful in the regex so we won't have any false positives. Can you write a test specifically for the regex pattern? It should have around 10 different paths, some positives and some negatives, and ensure the regex is accurate for all of them.

cool-RR (changes requested):
This change looks good to me. Can you add a test for the functionality?

pysnooper/tracer.py
luenk: - Regex changed to get a better match and changed to raw string
- Tests added to test the regex with different cases

cool-RR (changes requested):
Looks good, left one inline comment. Also, there should be a test for the actual usage (extracting source from the zip file).

luenk:
I added the test cases and changed the string/None handling. Please take a look to the changes.

tests/test_pysnooper.py
```diff
@@ -1914,3 +1916,147 @@ def f(x):
...
         f()
+
+
+def test_valid_zipfile():
+    with mini_toolbox.create_temp_folder(prefix='pysnooper') as folder, \
+                                    mini_toolbox.TempSysPathAdder(str(folder)):
+        module_name = 'my_valid_zip_module'
+        zip_name = 'valid.zip'
+        zip_base_path = 'ansible/modules'
+        python_file_path = folder / ('%s/%s/%s.py' % (zip_name, zip_base_path, module_name))
```
cool-RR, line 1927: Change to: 

```
python_file_path = folder / zip_name / zip_base_path / ('%s.py' % (module_name,))
```
Same for many other lines that have `folder /`.

tests/test_pysnooper.py
```diff
@@ -1914,3 +1916,147 @@ def f(x):
...
+        os.makedirs(folder / ('%s/%s' % (zip_name, zip_base_path)))
+        sys.path.insert(0, str(folder / (zip_name)))
```
cool-RR, line 1929: Every time you insert something to `sys.path`, do a `finally` to remove it after the test. I don't want it affecting other tests.

cool-RR: Also, you can remove parentheses here.

tests/test_pysnooper.py
```diff
@@ -1914,3 +1916,147 @@ def f(x):
...
+        content = textwrap.dedent(u'''
+            import pysnooper
+            @pysnooper.snoop(color=False)
+            def f(x):
+                return x
+        ''')
+        with python_file_path.open('w') as python_file:
```
cool-RR, line 1936: Check whether `write_text` works here instead of using `with`, I think it will.

tests/test_pysnooper.py
```diff
@@ -1914,3 +1916,147 @@ def f(x):
...
+            python_file.write(content)
+
+        module = importlib.import_module('%s.%s' % ('.'.join(zip_base_path.split('/')), module_name))
+
+        with zipfile.ZipFile(folder / 'foo_bar.zip', 'w') as myZipFile:
+            myZipFile.write(folder.joinpath('%s/%s/%s.py' % (zip_name, zip_base_path, module_name)), '%s/%s.py' % (zip_base_path, module_name,), zipfile.ZIP_DEFLATED)
+
+        python_file_path.unlink()
+        os.rename(folder.joinpath(zip_name), folder.joinpath('%s.delete' % (zip_name)),)
```
cool-RR, line 1992: Can you check whether `folder.joinpath(zip_name).rename(...)` would work instead of using `os.rename`? If so use everywhere.

cool-RR (changes requested):
Good tests, quite extensive.

You've got some long lines, please cut lines so no line is longer than 100 characters.

tests/test_pysnooper.py
luenk: Better readability - changed

tests/test_pysnooper.py
luenk: Fair - thanks for this hint

tests/test_pysnooper.py
luenk: It is working - changed

tests/test_pysnooper.py
luenk: It is working - changed. Thanks for the feedback

cool-RR (changes requested):
Good job. Note that you forgot some `breakpoint` lines in the code, please delete them. Then make sure your tests pass on Python 2.7 and at least one Python 3.x version. Then I'll merge them.

luenk:
Hi, 
that is interesting - perhaps there is also an error on Windows. 
I did the testing on new installed Linux environments - I will check this. 
Thanks for the feedback.
I am still working on the tests. At the moment I am working on the compatibility to python 2.7. 
It is not working with the imports in this way and there is also one additional error if a module was already imported.
Test case 1 and case 3 using the same package name.
I will come back to you with working tests. I need a little more time to fix these errors.
Thanks for testing,
Lukas

luenk:
Hi,
the tests are now python2 and Windows compatible.
Also Windows compatible regex and zip extraction was changed.

The test were running on:
platform darwin -- Python 2.7.18, pytest-4.6.11, py-1.11.0, pluggy-0.13.1
platform darwin -- Python 3.9.12, pytest-7.1.1, pluggy-1.0.0

platform linux2 -- Python 2.7.17, pytest-4.6.11, py-1.11.0, pluggy-0.13.1
platform linux -- Python 3.6.9, pytest-7.0.1, pluggy-1.0.0

platform win32 -- Python 2.7.18, pytest-4.6.11, py-1.11.0, pluggy-0.13.1
platform win32 -- Python 3.10.4, pytest-7.1.1, pluggy-1.0.0

Thanks for all your feedback,
Lukas

cool-RR:
Good job, merged. The `os.makedirs(p)` could be replaced by `p.mkdir(parents=True)`, but that shouldn't hold up the PR. I'll make a release.

luenk:
Thank you so much for merging and all your feedback.
I will take your feedback for future. 

Thanks and have a nice weekend,
Lukas

Issue #242, opened 2023-07-17
`return` event line missing when exception occurred


yantaozhao:
When exception occurred, the `return` event is not recorded in snoop log. This leaves an unbalanced `call-return` pairs.

**Expected behavior**: all ***event*** should properly be recorded even though exception happened.

demo code:
```python
import pysnooper

def bar(a, b):
    return a + b

def baz():
    try:
        x = 'hi' + 1  # exception here!
        return 1
    except:
        raise  # <- return event missing
    return -1

def foo():
    r = bar(2, 3)
    print(r)
    return r

@pysnooper.snoop(depth=2, normalize=True)
def main():
    foo()
    try:
        baz()
    except:
        pass

main()
```

![image](https://github.com/cool-RR/PySnooper/assets/9587405/3b16cb52-f78d-4f5c-863d-64a66c4d98c5)

In the above demo, `return` event in _line 13_ is not recorded like _line 20_ do. the `call-return` pair is broken.

cool-RR:
Which versions of PySnooper and Python? And what OS?

yantaozhao:
python 3.10, PySnooper 1.2.0, all platforms

cool-RR:
Ah, it's possible that I have just caused this bug with this most recent version. I'll look into it.

cool-RR:
If you're interested in writing a failing test case, that'll be helpful.

yantaozhao:
commit a PR to resolve this issue.

cool-RR:
Can you explain why there should be a `return` line when an exception is raised? Can you explain what practical problem it causes that this line doesn't exist?

yantaozhao:
> Can you explain why there should be a return line when an exception is raised?

No matter exception raised or not, all `return` event should be recorded because it really happened. It make things easy to write a log analyser as the `call-return` pairs are balanced.

> Can you explain what practical problem it causes that this line doesn't exist?

For log readable, consider below example:
```python
import pysnooper


def bar(a, b):
    return a + b


def baz():
    try:
        x = 'hi' + 1  # exception here!
        return 1
    except:
        raise
    finally:  # <-do something and output
        a=1
        b=2
        c=a+b
        foo()
    return -1


def foo():
    r = bar(2, 3)
    print(r)
    return [8]


@pysnooper.snoop(depth=9, normalize=True)
def main():
    try:
        baz()
    except:
        pass


main()
```

[...]

PR #243, opened 2023-07-17
Fix: missing return in exception event

yantaozhao:
Fix issue: https://github.com/cool-RR/PySnooper/issues/242


cool-RR:
Thanks for the quick PR. Could you please add a test that fails when this code is removed and succeeds with this code?

yantaozhao:
> Thanks for the quick PR. Could you please add a test that fails when this code is removed and succeeds with this code?

will do later when time is available.

yantaozhao:
> Thanks for the quick PR. Could you please add a test that fails when this code is removed and succeeds with this code?

updated

Issue #229, opened 2022-12-07
[feature request] Add command line color output for Windows


skjerns:
As requested in #1 , it would be nice to have colored output also for Windows, which should be possible, even though not as easily as in unix. 

I don't think I can currently implement it (as I'm also not using `PySnooper` very often).

cool-RR:
Cool, we'll wait for someone to volunteer to implement it. Note that it should be implemented without adding dependencies, so if we're using e.g. `colorama`, we'll need to copy-paste some of their code.

cool-RR:
If I remember correctly, the reason I didn't implement this when I implemented the same feature for Linux is that there are multiple shells you can run in Windows. (CMD, MSYS bash, Powershell, whatever Windows 10 has going on for running Linux.) I may have tried a solution on a couple of these and got inconsistent results. So figuring out a solution that works for all of these (or at least degrades gracefully) is the main difficulty here.

cool-RR:
The biggest priority in this feature is that whatever happens, we don't want to send an output with ugly shell escapes to a shell that won't interpret them correctly. That's the main disaster I'm trying to prevent.

SonGokussj4:
Just found out about PySnooper and it's a lovely library. I'm on Windows (Windows Terminal) and it's a shape it doesn't have colors.
If you want to have some compatibility between shells, what about additional library like `colorama` or something like that?
Second idea is to have `colors.conf` and let the user specify possible changes?

cool-RR:
Thank you. My second message in this thread answers part of your message. In any case, if you want to implement this, go ahead.

SonGokussj4:
Oh I apologize. I've skimmed through from 1st to 3rd and 4th message.

I see the code where you specify the colors. Will try to experiment a little bit. It's not my strong topic thought :-)
Anyway, have a great day.

P.S. Wouldn't it be possible to have something like like `pip install pysnooper[colorama]` which would be optional dependency or something like that? And then user can specify in `pysnooper.conf` if to use colorama or native shell escapes?

cool-RR:
> P.S. Wouldn't it be possible to have something like like `pip install pysnooper[colorama]` which would be optional dependency or something like that? And then user can specify in `pysnooper.conf` if to use colorama or native shell escapes?

It would be possible, but it means that for the rest of PySnooper's life I will have to confirm that these two branches work with each new release. The less moving parts, the less maintenace work I have to do.

Andrej730:
@cool-RR 

[...]

PR #258, opened 2025-04-10
Allow forcing color output on Windows using `color=True` #229

Andrej730:
As described in https://github.com/cool-RR/PySnooper/issues/229#issuecomment-2781296738, color output now is completely disabled on Windows. 

But since it's now used on Windows more widely, this PR removes this restriction and setting default value for `color` argument in `snoop` based on the current platform. Default value set to `True` on all other platforms besides Windows. This allows Windows users to access color output by using `snoop(color=True)`.

I guess, the only caveat of this PR is that default value now is platform-dependant and in theory someone can have a script with `snoop(color=True)` on Unix and when it's going to be executed on Windows, it will result in scary ANSI color codes. But it's all purely hypothetical since pysnooper is just a personal debugging tool.

### DIFF
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -225,7 +225,9 @@ class Tracer:
 
         @pysnooper.snoop(relative_time=True)
 
-    The output is colored for easy viewing by default, except on Windows.
+    The output is colored for easy viewing by default, except on Windows
+    (but can be enabled by setting `color=True`).
+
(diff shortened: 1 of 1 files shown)


cool-RR:
Good job. Add an empty line at the end of the two lines of documentation and I'll merge it.

Andrej730:
Sure.

PR #266, opened 2026-05-29
Guard against f_lasti == -1 when detecting exception returns

### LINKED ISSUE #260: Indexing frame.f_code.co_code at frame.f_lasti can use -1 (invalid) leading to wrong opcode or IndexError
## Summary

The tracer incorrectly reports "Call ended by exception" for normal function returns when `frame.f_lasti` is -1. This happens because the code accesses `frame.f_code.co_code[frame.f_lasti]` without checking if the index is valid.

In Python's tracing system, `frame.f_lasti` indicates the last instruction executed. However, this value can be -1 in certain cases (such as early function calls, some exception states, or in older Python versions).

When the code does `co_code[frame.f_lasti]` with `f_lasti = -1`, Python's negative indexing returns the last byte of the bytecode instead of throwing an error, thus,
this means we read the wrong opcode and make incorrect decisions about whether a function ended due to an exception.

**Current buggy code (in function trace() in tracer.py):**

```python
code_byte = frame.f_code.co_code[frame.f_lasti]  
if not isinstance(code_byte, int):
    code_byte = ord(code_byte)
ended_by_exception = (
    event == 'return'
    and arg is None
    and opcode.opname[code_byte] not in RETURN_OPCODES
)
```

It can cause a misleading debug output, thus, users may see "Call ended by exception" when functions return normally.

## Reproduction

I created a Python script with some tests that force this condition using mock frames:

```python
import unittest
import opcode
from unittest.mock import Mock

RETURN_OPCODES = {
    'RETURN_GENERATOR', 'RETURN_VALUE', 'RETURN_CONST',
    'INSTRUMENTED_RETURN_GENERATOR', 'INSTRUMENTED_RETURN_VALUE',
    'INSTRUMENTED_RETURN_CONST', 'YIELD_VALUE', 'INSTRUMENTED_YIELD_VALUE'
}


def buggy_ended_by_exception(frame, event, arg):
    """
    Original buggy logic: blindly indexes co_code[frame.f_lasti]
    """
    code_byte = frame.f_code.co_code[frame.f_lasti]
    if not isinstance(code_byte, int):
        code_byte = ord(code_byte)

    return (
        event == 'return'
        and arg is None
        and opcode.opname[code_byte] not in RETURN_OPCODES
    )


class TestNegativeLastiBug(unittest.TestCase):
    def test_negative_f_lasti_causes_wrong_opcode_read(self):
        """
        When f_lasti == -1, Python performs negative indexing and reads
        the *last* opcode, causing a normal return to be misclassified
        as an exception.
        """

        # last byte is NOP, which is not a return opcode:
        bytecode = bytes([
            opcode.opmap['LOAD_CONST'], 0,
            opcode.opmap['RETURN_VALUE'], 0,
            opcode.opmap['NOP'], 0,  
        ])

        mock_code = Mock()
        mock_code.co_code = bytecode

        frame = Mock()
        frame.f_code = mock_code
        frame.f_lasti = -1

        result = buggy_ended_by_exception(frame, 'return', None)

        # this test passes to prove the bug exists;
        # The buggy code returns True (wrong), so assertTrue passes:
        self.assertTrue(result, "Bug confirmed: normal return incorrectly classified as exception")


if __name__ == "__main__":
    unittest.main()
    
```

**Result:** Bug is confirmed; this is the test output in my terminal when running Python script I proposed above in Linux (Ubuntu) with Python 3.12 (note that a passing test ensure the bug exists):

```
.
----------------------------------------------------------------------
Ran 1 test in 0.000s

OK
```

## Proposed Fix

I also added the proposed fix in the Python script I did with tests. The code improvement simply checks `f_lasti >= 0` before accessing the bytecode:

```python
# determine the opcode at the last instruction safely. frame.f_lasti can be -1.
code_byte = None
if getattr(frame, 'f_lasti', -1) >= 0:
    try:
        code_byte = frame.f_code.co_code[frame.f_lasti]
    except Exception:
        # if anything odd happens accessing co_code, leave code_byte as None:
        code_byte = None
    else:
        if not isinstance(code_byte, int):
            code_byte = ord(code_byte)

opcode_name = None
if code_byte is not None:
    # opcode.opname may raise for invalid indices:
    try:
        opcode_name = opcode.opname[code_byte]
    except Exception:
        opcode_name = None

ended_by_exception = (
    event == 'return'
    and arg is None
    and (opcode_name is not None and opcode_name not in RETURN_OPCODES)
)
```

This improvement prevents confusing output when the bug does occur and may be more 'defensive' than the original code. I also replaced the proposed code for fixing in `tracer.py` and it didn't break previous tests belonging to the original test set of the project.

koriyoshi2041:
Fixes #260.

To distinguish a normal `return` event from one caused by an exception (both arrive as `event == 'return'` with `arg is None`), the tracer inspects the last executed opcode:

```python
code_byte = frame.f_code.co_code[frame.f_lasti]
...
ended_by_exception = (event == 'return' and arg is None
                      and opcode.opname[code_byte] not in RETURN_OPCODES)
```

`frame.f_lasti` can be `-1` (for example a frame that hasn't executed an instruction yet). `co_code[-1]` then reads the *last* byte of the bytecode — an unrelated opcode — and the result can flip `ended_by_exception` to `True`, so a normal return is printed as `Call ended by exception`.

This extracts the check into `call_ended_by_exception(frame, event, arg)` and returns `False` when `f_lasti < 0`, so the wrong opcode is never read.

Added `tests/test_ended_by_exception.py` with mock frames: `f_lasti == -1` is treated as a normal return; a return opcode → normal return; a non-return opcode → ended by exception; non-`return` events are never flagged. The `f_lasti == -1` case fails without the guard and passes with it; existing tracer tests still pass.

### DIFF
--- a/pysnooper/tracer.py
+++ b/pysnooper/tracer.py
@@ -29,6 +29,23 @@ RETURN_OPCODES = {
 }
 
 
+def call_ended_by_exception(frame, event, arg):
+    """Whether a 'return' event was caused by an exception, not a normal return.
+
+    On an exception the interpreter still emits a 'return' event with ``arg`` of
+    ``None``, so the last executed opcode is what tells them apart. A frame's
+    ``f_lasti`` can be ``-1`` (e.g. before any instruction has run); indexing
+    ``co_code[-1]`` would read the wrong opcode, so treat that as a normal
+    return. See https://github.com/cool-RR/PySnooper/issues/260.
+    """
+    if event != 'return' or arg is not None or frame.f_lasti < 0:
+        return False
+    code_byte = frame.f_code.co_code[frame.f_lasti]
+    if not isinstance(code_byte, int):
+        code_byte = ord(code_byte)
+    return opcode.opname[code_byte] not in RETURN_OPCODES
+
+
 def get_local_reprs(frame, watch=(), custom_repr=(), max_length=None, normalize=False):
     code = frame.f_code
     vars_order = (code.co_varnames + code.co_cellvars + code.co_freevars +
@@ -536,14 +553,7 @@ class Tracer:
         # If a call ends due to an exception, we still get a 'return' event
         # with arg = None. This seems to be the only way to tell the difference
         # https://stackoverflow.com/a/12800909/2482744
-        code_byte = frame.f_code.co_code[frame.f_lasti]
-        if not isinstance(code_byte, int):
-            code_byte = ord(code_byte)
-        ended_by_exception = (
-                event == 'return'
-                and arg is None
-                and opcode.opname[code_byte] not in RETURN_OPCODES
-        )
+        ended_by_exception = call_ended_by_exception(frame, event, arg)
 
         if ended_by_exception:
             self.write('{_FOREGROUND_RED}{indent}Call ended by exception{_STYLE_RESET_ALL}'.
--- a/tests/test_ended_by_exception.py
+++ b/tests/test_ended_by_exception.py
@@ -1,0 +1,41 @@
+"""Tests for call_ended_by_exception, the 'return'-via-exception detector.
+
+Regression test for https://github.com/cool-RR/PySnooper/issues/260: a frame's
+f_lasti can be -1, and co_code[-1] would read the wrong opcode and misreport a
+normal return as "Call ended by exception".
+"""
+
+import opcode
+from unittest.mock import Mock
+
+from pysnooper.tracer import call_ended_by_exception
+
+
+def _frame(co_code, f_lasti):
+    frame = Mock()
+    frame.f_lasti = f_lasti
+    frame.f_code.co_code = co_code
+    return frame
+
+
+def test_negative_f_lasti_is_not_an_exception():
+    # Last byte is a non-return opcode, so reading co_code[-1] (the old bug)
+    # would misreport an exception. With f_lasti == -1 it must be a normal
+    # return.
+    co_code = bytes([0, opcode.opmap["NOP"]])
+    assert call_ended_by_exception(_frame(co_code, -1), "return", None) is False
+
+
+def test_last_opcode_return_is_normal_return():
+    co_code = bytes([opcode.opmap["RETURN_VALUE"], 0])
+    assert call_ended_by_exception(_frame(co_code, 0), "return", None) is False
+
+
+def test_last_opcode_not_return_is_exception():
+    co_code = bytes([opcode.opmap["NOP"], 0])
+    assert call_ended_by_exception(_frame(co_code, 0), "return", None) is True
+
+
+def test_non_return_event_is_never_exception():
+    co_code = bytes([opcode.opmap["NOP"], 0])
+    assert call_ended_by_exception(_frame(co_code, 0), "call", None) is False
--- a/tests/test_ended_by_exception.py
+++ b/tests/test_ended_by_exception.py
@@ -5,9 +5,11 @@ f_lasti can be -1, and co_code[-1] would read the wrong opcode and misreport a
 normal return as "Call ended by exception".
 """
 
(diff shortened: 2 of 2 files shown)


cool-RR (commented):
Thanks @koriyoshi2041, good catch and a clean fix. I went and verified it carefully and it holds up.

**The bug is real on CPython 3.8 to 3.10 (and latent on 3.11+).**

The repro in #260 is mock-only, so I checked whether a real `return` event ever actually carries `f_lasti == -1`. Across normal returns, exceptions, and a pile of generator scenarios:

- 3.11 / 3.12 / 3.13: never happens. Purely latent.
- 3.8 / 3.9 / 3.10: it does. Closing (or `throw()`-ing into) an *unstarted* generator fires a `return` event whose frame has `f_lasti == -1`, which is exactly the case where `co_code[-1]` reads an unrelated byte.

And it surfaces through our own `depth>=2` path. On 3.8, master vs this PR for a plain `g = gen(5); g.close()`:

```
master                              this PR
  exception 12 def gen(n):            exception 12 def gen(n):
  Call ended by exception      -->    return    12 def gen(n):
                                       Return value:.. None
```

master's "exception" verdict there is just luck of the last bytecode byte, not detection, so I'm glad to see it go.

**Verification**
- Full suite passes on 3.8, 3.12, 3.13 (plus the 4 new unit tests).
- Behavior is identical to master except the `f_lasti < 0` case, which is the fix. It also can't hide a genuine error: with nothing executed in the frame the only way out is an injected exception, which still gets its own `exception` event.
- Bonus: moving the opcode read behind the `event == 'return'` check skips it on every line event, which is the hot path.

**How others handle it, for reference**
- `python-hunter` guards the same way before indexing: `len(co_code) > f_lasti >= 0`. Same instinct as this PR.
- `snoop` has the identical unguarded `co_code[f_lasti]` and nobody's hit it, which matches the "latent on modern CPython" finding.
- stdlib `bdb`/`pdb` sidestep it entirely by routing on the event string and never reading opcodes.

**Two small things**
1. The new tests are unit-only (mocked frames). Fair, since the bug isn't reachable on CI Python, but I'd like a real integration test alongside them (below). It passes on all versions and would have failed on master for 3.8 to 3.10.
2. `tests/test_ended_by_exception.py` imports `unittest.mock`, which is Py3-only, while `setup.py` still lists 2.7. That 2.7 classifier is vestigial and I'll drop it; just flagging so the test file isn't what trips over it.

**Suggested integration test** (verified: passes on this PR for 3.8/3.10/3.12/3.13, fails on master for 3.8/3.10):

```python
def test_unstarted_generator_close_is_not_an_exception():
    # Regression for #260. On CPython 3.8-3.10, closing an unstarted generator
    # fires a 'return' event whose frame has f_lasti == -1; the old code read
    # co_code[-1] (an unrelated opcode) and mislabeled this normal close as
    # "Call ended by exception". Passes on all versions; only catches the bug
    # on 3.8-3.10, where it is reachable.
    string_io = io.StringIO()

    def gen(n):
        for i in range(n):
            yield i

    @pysnooper.snoop(string_io, depth=2, color=False)
    def main():
        g = gen(5)
        g.close()
        return 42

    assert main() == 42
    output = string_io.getvalue()
    assert 'Call ended by exception' not in output
    assert 'Return value:.. 42' in output
```

(needs `import io` and `import pysnooper` at the top of the file.)

Add that and I'm happy to merge. Thanks again!

koriyoshi2041:
Added the integration regression test you suggested and pushed it as `393eb37`.

Checked locally with:

```text
uv run --with pytest python -m pytest tests/test_ended_by_exception.py -q
# 5 passed
```

cool-RR:
Good job, thank you.

PR #267, opened 2026-07-04
report variable changes from the last line in a with block

### LINKED ISSUE #237: No variable output for the last line
Hi, I found that the variable, new or modified, in the last line will not be shown. Below is the same example with the `print` removed:
```
import pysnooper
import random

def foo():
    lst = []
    for i in range(10):

    with pysnooper.snoop():
        lower = min(lst)
        upper = max(lst)
        mid = (lower + upper) / 2

foo()
```
which outputs something like:
```
New var:....... lst = [379, 800, 331, 57, 210, 524, 525, 837, 118, 810]
New var:....... i = 9
22:46:58.874665 line        10         lower = min(lst)
New var:....... lower = 57
22:46:58.876660 line        11         upper = max(lst)
New var:....... upper = 837
22:46:58.876660 line        12         mid = (lower + upper) / 2
Elapsed time: 00:00:00.003000
```
Where the variable `mid` was not printed. Please let me know if this is expected.

CodingSelim:
Fixes #237.

Variable changes are diffed and printed on the *next* trace event. When `snoop()` is used as a context manager, the last line of the block has no next event before `__exit__`, so its new/modified variables were never printed. On Python 3.10+ an extra block-exit line event happens to flush them, which is why it reproduced on 3.9 but not later.

Flush the pending changes for the traced frame in `__exit__` too (a no-op when a later event already reported them, so no duplicates on 3.10+). Pulled the variable-reporting logic into a small helper shared by `trace` and `__exit__`.

Updated the existing with-block tests: the last-in-block variable is now reported on all versions (dropped the `min_python_version=(3, 10)` guard on those `VariableEntry`s), while the 3.10+ block-exit line entry stays guarded.

Verified locally on 3.11; relying on CI for the older versions where the original bug reproduces.


cool-RR (changes requested):
Thanks for working on this. This is a real bug, and the basic direction is right.

I tested this PR in Bubblewrap on CPython 3.8 through 3.14 and PyPy 3.10. Your changed expectations fail against master on Python 3.8 and 3.9, and pass with your patch, so the fix for the reported multiline case is real.

There is one blocker before I can merge it.

On Python 3.10 and newer, Python already sends a line event when leaving the `with` block. That event reports the final variable state. `__exit__` then calls `_report_variable_changes` again, so every local repr, watch expression, and custom-repr predicate is evaluated twice.

This isn't a harmless no-op. I tested two concrete cases:

1. An object whose `__repr__` changes on every call. Master reports it once as a new variable. This PR calls it twice and prints a fake `Modified var` line.
2. A `custom_repr` predicate that succeeds once and raises on its second call. Master completes normally. This PR raises `RuntimeError` from `__exit__` and changes the behavior of the program being debugged.

Please change the implementation so the exit flush happens only when the final state has not already been captured. I'd prefer tracking whether the context-exit line event occurred, rather than checking `sys.version_info`, so this follows the interpreter's actual behavior.

Please also add dedicated regression tests for this bug. The current changes indirectly loosen three existing expectations, but:

- Two use a bare `VariableEntry()`, which accepts any variable name, value, or stage.
- All three test a newly created local. None tests a modified existing local, even though the issue and PR claim to fix both.
- None detects the duplicate evaluation described above.

A good regression test would initialize an existing variable, then make the final body line both modify that variable and create a new one. Assert the exact stage, name, value, order, and that each line appears once. Run the same test on Python 3.9 and a current Python.

Please add another test with a counted watch or custom-repr predicate, proving that leaving the block does not evaluate it redundantly on current Python.

I also found that the valid one-line form is still completely silent except for elapsed time:

```python
with pysnooper.snoop(output): answer = 42
```

This happens because there is no trace event after `__enter__`, so `frame_to_local_reprs` never gets an entry and the new guard skips the flush. Please handle this case too and add a test for it.

There is no functioning CI on the repository right now. After updating the PR, please run the full suite on Python 3.9 and one current Python version and paste the results here.

Once you've pushed those changes, I'll test and review it again. Thanks.

CodingSelim:
good catch on the double eval, fixed. the exit flush now only runs when the block-exit line event didn't already fire, and i track that from the actual event (a line event landing back on the `with` line, stored in `frame_to_with_line`) instead of `sys.version_info`. so on 3.10+ `__exit__` sees it was already flushed and does nothing, no second repr/watch/custom_repr call. both your cases check out now: the mutating `__repr__` prints once, and the `custom_repr` that raises on the 2nd call no longer raises.

the one-line `with snoop(): x = 1` form goes through the same path, it produces no trace event at all so `frame_to_with_line` never gets marked and `__exit__` does the flush, so it reports the var now.

added three regression tests:
- a final line that both modifies an existing variable and creates a new one, asserting exact stage/name/value/order with each line once, runs on 3.9 and 3.10+ (the only difference is the extra with-line entry on 3.10+ which stays guarded)
- the one-line body
- a counted `custom_repr` proving it's evaluated exactly once. this one fails with `assert 2 == 1` against the previous version, so it pins the double eval.

ran the full suite:

```
$ python --version && python -m pytest tests/test_pysnooper.py -q
Python 3.9.25
74 passed, 2 skipped

$ python --version && python -m pytest tests/test_pysnooper.py -q
Python 3.14.2
76 passed
```

thanks for the detailed review.

cool-RR (changes requested):
Thanks — the changes you made do fix the cases from my first review. The final modified/new variables are now asserted precisely, the one-line form works, and normal exits no longer double-evaluate on current Python.

I reran the complete test file in Bubblewrap on CPython 3.8, 3.9, 3.10, 3.11, 3.12, 3.13, 3.14, and PyPy 3.10. Everything passes. I also transplanted the three new tests onto master and the previous PR head; they go red in the intended places, so those tests are meaningful.

There are still two blockers on the old Python versions this PR is meant to fix.

First, exceptional exits still evaluate variables redundantly on CPython 3.8 and 3.9. Those versions have already captured the post-statement state through the exception/cleanup trace events, but `frame_to_with_line` remains set because it is cleared only by a line event whose line number equals the stored `with` line. `__exit__` consequently takes another snapshot.

With a counted custom-repr callback, exact base -> PR calls are:

- Python 3.8: 2 -> 3
- Python 3.9: 3 -> 4
- Python 3.10, 3.12, 3.14, and PyPy 3.10: unchanged

This can change the program being debugged. I made a custom-repr predicate raise on precisely the added call. On base, both old Pythons propagate the user's original `ValueError`. On this head, both replace it with the predicate's `RuntimeError`. Because that happens inside `_report_variable_changes`, the remaining per-frame state is not cleaned up either.

Please add an exceptional-exit regression test on Python 3.8/3.9 which proves all three things: no added callback evaluation, the original exception survives, and tracer state is cleaned up. Please also make the cleanup exception-safe rather than allowing a callback failure to strand frame references.

Second, `calling_frame.f_lineno` in `__enter__` is not reliably the `with` statement's line on Python 3.8/3.9. When the same source-level statement executes again in a loop, it can still be the previous body line. The body's pre-execution line event then accidentally clears `frame_to_with_line`, so `__exit__` skips the flush that was actually needed.

Minimal reproduction:

```python
for i in range(3):
    with pysnooper.snoop(output):
        x = i
```

On this head under both Python 3.8 and 3.9, the first iteration reports final `x = 0`, but the next two report only the stale starting values `x = 0` and `x = 1`. Final `x = 1` and `x = 2` are never reported. Python 3.10+ behaves correctly.

Please add a regression test that repeatedly executes the same source-level `with` statement and asserts every iteration's final value. The implementation should track whether a post-body snapshot actually happened without assuming that `f_lineno` at `__enter__` identifies the `with` line.

Once those two paths are fixed and tested on Python 3.9 plus a current Python, I'll rerun the matrix. The ordinary normal-exit behavior in this revision looks good.

CodingSelim:
both fixed.

on the exceptional exit, `__exit__` now skips the flush entirely when there's an exception (`exc_type is not None`), since the exception/return events already captured the final state on those versions. so no extra callback eval, the user's exception propagates untouched, and i moved the per-frame cleanup into a `finally` so a failing callback can't strand frame refs.

for the loop case i stopped trusting `f_lineno` at `__enter__` and resolve the `with` line from `f_lasti` via `dis.findlinestarts`. the flush is also no longer gated on matching that line, it's driven by a pending flag that gets cleared by the block-exit line event or any non-line event (exception/return), so exceptional exits and re-run statements both clear it correctly. your loop repro now reports all three finals on 3.8/3.9 (x = 0 new, 1 and 2 modified).

added two tests:
- a `with` block reused across a loop, asserting each iteration's final value
- exceptional exit, asserting `__exit__` takes no snapshot, the original `ValueError` propagates, and every per-frame dict is cleaned up

both go red on the previous head in the right spots.

full suite:

```
$ python --version && python -m pytest tests/test_pysnooper.py -q
Python 3.9.25
76 passed, 2 skipped

$ python --version && python -m pytest tests/test_pysnooper.py -q
Python 3.14.2
78 passed
```

(also green on 3.8.20 and 3.11.) thanks again.

cool-RR (approved):
Thanks for sticking with this through the edge cases and for responding so carefully to both rounds of review.

I retested this exact head in Bubblewrap on Python 2.7, CPython 3.4 through 3.14, and PyPy 3.10. The full suite is green throughout. I also reran independent probes for the original last-line bug, one-line blocks, repeated `with` statements, exceptional exits, callback evaluation, cleanup, return, break, and continue paths.

The original issue is fixed, both review blockers are resolved, and I found no remaining regression. This closes #237 cleanly. Thanks for the solid work.

CodingSelim:
appreciate you running it across the whole matrix down to 2.7 and probing the break/continue/return paths too, that's way more coverage than i could hit locally. glad it's sorted.

